PR #3, opened 2016-06-20
Add file based dmclock configuration parameter functionality

bspark8:
When executing dmc_sim, configuration filename will be received as parameter so that we could execute multiple tests easily.
For example, for new test with new parameters(like server count, client iops... etc) we just add new configuration file without recompiling.
Moreover we can replay the test just with the configuration file we want to test.

**- Usage**
**1. ./sim/dmc_sim -c/--conf ./dmc_sim.conf (dmclockparameter configuration file)**
**2. dmc_sim.conf(example)**

**[global]**
**server_count= 1**                              _// the total number of servers_
**client_count= 3**                                _// the total number of clients_
**server_types= 1**                              _// types of servers (Ex. server_count=100, server_types=3 >> server i(0~99)‘s type = i% server_types)_
**client_types= 3**                                _// types of clients_
**server_random_selection= false**  _// client’s seed or random server selection way_
**server_soft_limit= false**                  _// allow limit break_

**[client.0]**                                            _// defined as the number of client types_
**client_wait= 0**                                   _// starting waiting time_
**client_total_ops= 10000**                 _// the total number of request per client_
**client_server_select_range= 1**      _// client server selection range_
**client_iops_goal= 1000**                  _// client request issue capacity_
**client_outstanding_ops= 32**          _// client concurrent I/Os_
**client_reservation= 0.0**                   _// tag reservation_
**client_limit= 0.0**                                _// tag limit_
**client_weight= 1.0**                           _// tag weight_

**[client.1]**
**client_wait= 5**
**client_total_ops= 10000**
**client_server_select_range= 1**
**client_iops_goal= 1000**
**client_outstanding_ops= 32**
**client_reservation= 0.0**
**client_limit= 200.0**
**client_weight= 1.0**

**[client.2]**
**client_wait= 10**
**client_total_ops= 10000**
**client_server_select_range= 1**
**client_iops_goal= 1000**
**client_outstanding_ops= 32**
**client_reservation= 0.0**
**client_limit= 250.0**
**client_weight= 2.0**

**[server.0]**                                              _// defined as the number of server types_
**server_iops= 800**                                _// server I/O processing capacity_
**server_threads= 1**                              _// the number of worker threads per server_

### DIFF
--- a/dmc_sim.conf
+++ b/dmc_sim.conf
@@ -1,0 +1,41 @@
+[global]
+server_count = 1
+client_count = 3
+server_types = 1
+client_types = 3
+server_random_selection = false
+server_soft_limit = false
+
+[client.0]
+client_wait = 0
+client_total_ops = 2000
+client_server_select_range = 1
+client_iops_goal = 200
+client_outstanding_ops = 32
+client_reservation = 0.0
+client_limit = 0.0
+client_weight = 1.0
+
+[client.1]
+client_wait = 5
+client_total_ops = 2000
+client_server_select_range = 1
+client_iops_goal = 200
+client_outstanding_ops = 32
+client_reservation = 0.0
+client_limit = 40.0
+client_weight = 1.0
+
+[client.2]
+client_wait = 10
+client_total_ops = 2000
+client_server_select_range = 1
+client_iops_goal = 200
+client_outstanding_ops = 32
+client_reservation = 0.0
+client_limit = 50.0
+client_weight = 2.0
+
+[server.0]
+server_iops = 160
+server_threads = 1
--- a/dmc_sim.conf
+++ b/dmc_sim.conf
@@ -1,12 +1,11 @@
 [global]
-server_count = 1
-client_count = 3
-server_types = 1
-client_types = 3
+server_groups = 1
+client_groups = 3
 server_random_selection = false
 server_soft_limit = false
 
 [client.0]
+client_count = 1
 client_wait = 0
 client_total_ops = 2000
 client_server_select_range = 1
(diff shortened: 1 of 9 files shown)


ivancich:
One of the tests that has been useful for me is to have 99 clients operate one way and then 1 client operate differently. Am I correct that if you define multiple types of clients, they're fully interleaved? That is if there are 100 clients and 3 types, 34 clients will be of type 0, 33 of type 1, and 33 of type 2?

What would it take to allow more control over this distribution? Perhaps client_count could be defined in each client type section. This would complicate the logic for the client info function since it'd have to consider how many of each client type there were. And rather than interleaving them, it'd probably better to assign them in contiguous blocks. So client type 0 would be clients 0...i. client type 1 would be clients i+1...j. Etc.

I think we should use a similar system for servers as well, where the counts are defined per type.

bspark8:
I totally agree with you. I changed configuration file rule so that each client count can be defined in each client group (before name type) section. Also, server as well.
And rather than interleaving them, assign them in contiguous blocks.

**Usage**
**1. ./sim/dmc_sim (use default original parameter)**
or
**2. ./sim/dmc_sim -c/--conf ./dmc_sim.conf (dmclock parameter configuration file)**

**Default original parameter example**
[global]
**server_groups = 1**
**client_groups = 2**
server_soft_limit = true

[client.0]
**client_count = 99**
**client_wait = 0**
client_total_ops = 1000
client_server_select_range = 10
client_iops_goal = 50
client_outstanding_ops = 100
client_reservation = 20.0
client_limit = 60.0
client_weight = 1.0

[client.1]
**client_count = 1**
**client_wait = 10**
client_total_ops = 1000
client_server_select_range = 10
client_iops_goal = 50
client_outstanding_ops = 100
client_reservation = 20.0
client_limit = 60.0
client_weight = 1.0

[server.0]
**server_count = 100**
server_iops = 40
server_threads = 1

Thank you.

ivancich:
Thank you for that contribution!

PR #4, opened 2016-06-24
Defer tag calculation with up-to-date delta/rho

TaewoongKim:
We found clients’ IOPS were up&down and not stable under dmc_sim’s default parameter(server/client count, iops, … etc.). We analyzed this situation and concluded that dmclock has a potential problem in the special case. The problem comes out if there are too many requests in the client request queue on a server.

The main reason is that the tag was calculated too early. Requests’ tag are calculated on received time on server side with request’s rho&delta(the client’s status) right before pushing in the request queue.
If the request queue is too long, the tag value is not reflect the current client status when the request is being scheduled because too long time was passed during serving previous requests. Consequently, the tag value calculated based on too old rho&delta induces unstable QoS.

To improve this, we changed tag calculation timing. After changing calculate time to right before pushing a request to heaps(not before pushing client request queue), we could get improved QoS result. The changes was not much. Just added current rho&delta for each clientRec and kept it up-to-date.(update with every arriving request’s rho&delta). Before pushing tag to heaps, we calculated tag based on current rho/delta(not based on the request’s rho/delta) so that tag’s value could reflect recent status.

**Before**
image

**After modification**
image

**Test parameters**

    // server params
    const uint server_count = 100;
    const uint server_iops = 40;
    const uint server_threads = 1;
    const bool server_soft_limit = true;

    // client params
    const uint client_total_ops = 2000; // for longer test
    const uint client_count = 100;
    const uint client_server_select_range = 10;
    const uint client_wait_count = 1;
    const uint client_iops_goal = 50;
    const uint client_outstanding_ops = 100;
    const std::chrono::seconds client_wait(10);

    // client info
    const double client_reservation = 20.0;
    const double client_limit = 60.0;
    const double client_weight = 1.0;

### DIFF
--- a/src/dmclock_server.h
+++ b/src/dmclock_server.h
@@ -246,6 +246,8 @@ namespace crimson {
 	ClientInfo            info;
 	bool                  idle;
 	Counter               last_tick;
+	uint32_t              cur_rho;
+	uint32_t              cur_delta;
 
 	ClientRec(C _client,
 		  const ClientInfo& _info,
@@ -254,7 +256,9 @@ namespace crimson {
 	  prev_tag(0.0, 0.0, 0.0),
 	  info(_info),
 	  idle(true),
-	  last_tick(current_tick)
+	  last_tick(current_tick),
+	  cur_rho(1),
+	  cur_delta(1)
 	{
 	  // empty
 	}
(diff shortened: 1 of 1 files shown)


isis1212:
Yep, that makes sense. If there are problems, we can simply test original behavior to know whether the modification produces it or not.  I added some "#ifdef"s simply as you requested.

TaewoongKim:
Yep, that makes sense. If there are problems, we can simply test original behavior to know whether the modification produces it or not. I added some "#ifdef"s simply as you requested.

ivancich:
@TaewoongKim : It turned out this PR caused a unit test to fail. I thought I'd checked it. I looked into it and it seemed the problem was the time used when a tag was finally calculated. In the original code, the time passed into the request was used; this was the time of arrival or a time specified by a unit test. In this PR the time of tag calculation was always used.

I believe I fixed it in commit 2dba0a11dd1bf6fdb0ffdb0841d2cae9ed66b0a1. Would you mind looking at my fix and seeing whether it undermines your improvements? Thank you!

TaewoongKim:
@ivancich : I commented about the commit 2dba0a1 at the commit page. Thank you~

PR #5, opened 2016-07-11
Large client_total_ops

TaewoongKim:
I recently tested large numbers for config parameters for testing high performance device configuration. But, client_total_ops was uint16_t  type. Because it was too small, test didn't work properly.
In this PR, I changed some variables' type from uint16_t to uin32_t. And I fixed output format to print large number properly. Because large numbers with long digits were concatenated with each others without delimiter. It was hard to read the number. Therefore, I added single space between output numbers. But small numbers will be printed same as previous result output because I reduced 'data_w' by 1 for compensation about one added blank.

```
#Test Configuration
[global]
server_groups = 1
client_groups = 1
server_soft_limit = false

[client.0]
client_count = 5
client_wait = 0
client_total_ops = 150000
client_server_select_range = 10
client_iops_goal = 15000
client_outstanding_ops = 64
client_reservation = 20.0
client_limit = 20000.0
client_weight = 1.0

[server.0]
server_count = 16
server_iops = 5500
server_threads = 10
```

### DIFF
--- a/sim/src/sim_client.h
+++ b/sim/src/sim_client.h
@@ -36,7 +36,7 @@ namespace crimson {
       union {
 	std::chrono::milliseconds wait_time;
 	struct {
-	  uint16_t count;
+	  uint32_t count;
 	  std::chrono::microseconds time_bw_reqs;
 	  uint16_t max_outstanding;
 	} req_params;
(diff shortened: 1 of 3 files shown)


ivancich:
Thank you!

PR #7, opened 2016-07-18
Indirect Intrusive heap base, and USE_SIMPLE_HEAP flag in dmclock_server

smbillah:
Added a base-class named indirect_intrusive_base which is the parent of all indirect_intrusive heap implementation. The goal is to dynamically change heap data structure (for example, linear heap, or regular heap) in dmclock_server depending on the number of clients present at a time. 
    
Also, added USE_SIMPLE_HEAP flag to turn on simple linear heap (named ndirect_intrusive_linear_heap), that uses linear search, i.e., with O(num_of_clients) complexity. 
This simple heap can serve as a baseline for all future heap-implementation.

### DIFF
--- a/sim/src/config.h
+++ b/sim/src/config.h
@@ -73,13 +73,15 @@ namespace crimson {
       uint server_count;
       uint server_iops;
       uint server_threads;
+      bool server_use_heap;
 
       srv_group_t(uint _server_count = 100,
 		  uint _server_iops = 40,
 		  uint _server_threads = 1) :
 	server_count(_server_count),
 	server_iops(_server_iops),
-	server_threads(_server_threads)
+	server_threads(_server_threads),
+	server_use_heap (true)
       {
 	// empty
       }
@@ -89,7 +91,8 @@ namespace crimson {
 	out <<
 	  "server_count = " << srv_group.server_count << "\n" <<
 	  "server_iops = " << srv_group.server_iops << "\n" <<
-	  "server_threads = " << srv_group.server_threads;
+	  "server_threads = " << srv_group.server_threads << "\n" <<
+	  "server_use_heap = " << srv_group.server_use_heap;
 	return out;
       }
     }; // class srv_group_t
--- a/support/src/indirect_intrusive_base.h
+++ b/support/src/indirect_intrusive_base.h
@@ -1,0 +1,370 @@
+// -*- mode:C++; tab-width:8; c-basic-offset:2; indent-tabs-mode:t -*-
+// vim: ts=8 sw=2 smarttab
+
+/*
+ * Copyright (C) 2016 Red Hat Inc.
+ */
+
+
+#pragma once
+
+#include <memory>
+#include <vector>
+#include <string>
+#include <iostream>
+#include <functional>
+
+#include "assert.h"
+
+
+namespace crimson {
+  using IndIntruHeapData = size_t;
+
+  /* T is the ultimate data that's being stored in the heap, although
+   *   through indirection.
+   *
+   * I is the indirect type that will actually be stored in the heap
+   *   and that must allow dereferencing (via operator*) to yield a
+   *   T&.
+   *
+   * C is functor when given two T&'s will return true if the first
+   *   must precede the second.
+   *
+   * heap_info is a data member pointer as to where the heap data in T
+   * is stored.
+   */
+  template<typename I, typename T, IndIntruHeapData T::*heap_info,  typename C>
+  class IndIntruBase {
+
+    static_assert(
+      std::is_same<T,typename std::pointer_traits<I>::element_type>::value,
+      "class I must resolve to class T by indirection (pointer dereference)");
+
+    static_assert(
+      std::is_same<bool,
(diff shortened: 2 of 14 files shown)


support/src/indirect_intrusive_heap.h
```diff
@@ -33,9 +26,9 @@ namespace crimson {
    *
    * heap_info is a data member pointer as to where the heap data in T
    * is stored.
-   */
+   */ //public Base<int, short, double>
```
ivancich: What's this commented out definition of Base for, and why such a generic name?

smbillah: My bad. ... that line was part of the experiment which I should've removed.

ivancich: No problem.

sim/src/config.h
```diff
@@ -73,13 +73,15 @@ namespace crimson {
       uint server_count;
       uint server_iops;
       uint server_threads;
+      bool server_use_heap;
```
ivancich: Did I miss the role of this variable?

support/test/CMakeLists.txt
```diff
@@ -19,12 +19,12 @@ set_source_files_properties(${srcs} ${test_srcs}
...
 add_executable(data-struct-tests EXCLUDE_FROM_ALL ${test_srcs})
 
-set_target_properties(data-struct-tests
+set_target_properties(data-struct-tests 
   PROPERTIES
   RUNTIME_OUTPUT_DIRECTORY ../../test)
 
 target_link_libraries(data-struct-tests
-  PUBLIC pthread ${GTEST_LIBRARY} ${GTEST_MAIN_LIBRARY})
+  PUBLIC ${GTEST_LIBRARY} ${GTEST_MAIN_LIBRARY} pthread )
```
ivancich: Out of curiosity, did the re-ordering solve something?

support/src/indirect_intrusive_base.h
```diff
@@ -0,0 +1,27 @@
...
+#include <functional>
+
+#include "assert.h"
+
+
+namespace crimson {
+  using IndIntruHeapData = size_t;
+
+  template<typename I, typename T, IndIntruHeapData T::*heap_info,  typename ... Cs>
+  class IndIntruBase {
```
ivancich: I'm unclear as to the role of this class. As best I can tell it combines a bunch of **_#include_**s and a **_using_**.

ivancich:
Please rebase on the latest ceph/dmclock master.

ivancich:
I don't think it's correct to call your data structure a _heap_. Does it
satisfy the heap property <see:
[Wikipedia article on heap data structure](https://en.wikipedia.org/wiki/Heap_(data_structure)) >?

support/src/indirect_intrusive_linear_heap.h
```diff
@@ -0,0 +1,298 @@
...
+    CmpR                     cmp_resv;
+    CmpP                     cmp_ready;
+    CmpL                     cmp_limit;
+
+  protected:
+    using index_t = IndIntruHeapData;
+
+    std::vector<I>           data;
+    index_t                  count;
+    T                        *resv;
```
ivancich: Are these three variables used so that identifying the next reservation/limit/ready and then popping it doesn't require two searches? How different would the code be if they weren't there? And a related comment is this is not a generic data structure -- it has a bunch of information about dmclock built into it. I don't know if that's necessarily bad, but I'm somewhat certain that the indirect intrusive heap is generic.

sim/src/config.h
smbillah: No. This variable has no usage as of now, but will come handy (during testing, debugging) if we want dmclock to switch between two data-structure (heap, linear) dynamically.

support/src/indirect_intrusive_base.h
smbillah: This class will come handy if we decide to put virtual heap methods such as top(), empty(), adjust(), etc., plus iterator, const iterator.

support/src/indirect_intrusive_linear_heap.h
smbillah: Are these three variables used so that identifying the next reservation/limit/ready and then popping it doesn't require two searches? 
-- yes.   
How different would the code be if they weren't there? -- need to use 3 instances of this linear_heap (or something more appropriately named) structure, each of which will contain one comparator (in contrast to one instance having 3 comparators).

support/test/CMakeLists.txt
smbillah: http://stackoverflow.com/questions/10315016/error-during-making-gtest

ivancich: Thanks.

support/src/indirect_intrusive_base.h
ivancich: This commit should be clean and not contain code for a possible future version.

sim/src/config.h
ivancich: This commit should be clean and not contain code for a possible future version.

ivancich:
Please rebase this on latest master (see request of 3 days ago).

PR #8, opened 2016-07-20
Change limit heap to use adjust() instead of demote() after request pop.

TaewoongKim:
The selected client is not the top of all heaps. So, some heaps need
promotion not demotion. The effect of this change can be seen at the below configuration clearly.

```
[global]                       
server_groups = 1              
client_groups = 2              
server_soft_limit = true       
                               
[client.0]                     
client_count = 1               
client_wait = 0                
client_total_ops = 1000        
client_server_select_range = 4 
client_iops_goal = 60          
client_outstanding_ops = 100   
client_reservation = 20.0      
client_limit = 70.0            
client_weight = 1.0            
                               
[client.1]                     
client_count = 1               
client_wait = 0                
client_total_ops = 8000        
client_server_select_range = 4 
client_iops_goal = 800         
client_outstanding_ops = 100   
client_reservation = 20.0      
client_limit = 1000.0          
client_weight = 200.0          
                               
[server.0]                     
server_count = 4               
server_iops = 80               
server_threads = 1             

```

### DIFF
--- a/src/dmclock_server.h
+++ b/src/dmclock_server.h
@@ -775,7 +775,7 @@ namespace crimson {
 #endif
 
 	resv_heap.demote(top);
-	limit_heap.demote(top);
+	limit_heap.adjust(top);
 #if USE_PROP_HEAP
 	prop_heap.demote(top);
 #endif


ivancich:
This is a very good catch; thank you!

However my thinking is that only the **limit_heap** needs to be switched from **demote** to **adjust**. I believe we're guaranteed that each of the tag values (reservation, proportion, and limit) will either get larger or stay the same, which in general will mean the next request tag can only move down or not move. The one **exception** is with the **limit_heap** in which the **ready** flag is also used for ordering. Tags with ready=true are lower in the heap than tags with ready=false. However the next item in the heap could have ready=true, allowing it to be promoted rather than demoted.

When I try the code where the change is only applied to **limit_heap** I get essentially the same results as to when your entire commit is included.

Further, I checked my assumptions by inserting these four asserts into the top.has_request() conditional.

	  assert(next_first.tag.reservation >= first.tag.reservation);
	  assert(next_first.tag.proportion >= first.tag.proportion);
	  assert(next_first.tag.limit >= first.tag.limit);
	  assert(!next_first.tag.ready || first.tag.ready);

There was never a conflict with any of the assertions.

Let me know what you think. Perhaps you have a counter-example. Thank you!

TaewoongKim:
You are right. I was misunderstanding about resv/ready heap. They  just need demotion because requests' reservation/proportion tag will be grow always and ready flag=false(initial value) make requests go down in heaps except limit heap. I will change the code and re-push as your comment.

ivancich:
Thank you for that bug fix. I really appreciate your contributions to the project.

TaewoongKim:
You're welcome. I just wish Ceph QoS developing is completed as soon as possible. My little helps will be continued until that. :)

PR #10, opened 2016-08-01
K way heap

smbillah:
A proper K-value in generic k-way heap could save up-to 100-200 ns on every heap operation.


smbillah:
The first purple bar stands for existing heap-implementation. The subsequent bars stand for heaps with K value from 1 to 10.
[client_all.pdf](https://github.com/ceph/dmclock/files/394057/client_all.pdf)

ivancich:
What is the setting of server_select_range in these tests? I'm trying to figure out why when server count is 100, it runs faster when there are 100 clients than only 10 clients.

benchmark/test
```diff
@@ -0,0 +1,31 @@
+
+file_name:1:configs/dmc_sim_8_6.conf
```
ivancich: This looks like it's an output file. If so, it should not be in the commit.

support/src/indirect_intrusive_heap.h
```diff
@@ -33,8 +33,10 @@ namespace crimson {
    *
    * heap_info is a data member pointer as to where the heap data in T
    * is stored.
+   *
+   * K is the branching factor of the heap, default is 2 (binary heap).
    */
-  template<typename I, typename T, IndIntruHeapData T::*heap_info, typename C>
+  template<typename I, typename T, IndIntruHeapData T::*heap_info, typename C, uint K = 2>
```
ivancich: I'd like to think about whether having K be a template argumetn is the best way. It could just as easily be determined at runtime, which then lends itself to being set in the configuration file. What are the arguments for leaving it as a template argument?

support/src/indirect_intrusive_heap.h
```diff
@@ -196,10 +198,10 @@ namespace crimson {
     IndIntruHeap() :
       count(0)
     {
-      // empty
+      assert(K > 0);
```
ivancich: If it is left as a template argument, then I think we should see if a static (compile-time) assertion could work. If we end up making it a runtime parameter, then an assert would be fine.

Also, why not K > 1? If K is 1, then we're keeping the structure fully sorted using something like an incremental bubble sort, which isn't very efficient.

ivancich:
I think the k-way heap needs some good unit tests.

ivancich:
I don't think the PDF should be left in the commit, especially if it can be recreated by the tooling in the benchmark directory.

benchmark/test
smbillah: It's a sample config file that is required to run the benchmark script successfully. However, there are 3 output files named "test", "test.dat", and "test.pdf" which I removed from the commit.

support/src/indirect_intrusive_heap.h
smbillah: Here are the pros and cons of using K as a template argument:
pros:
-- minimal code change (since K is static)

cons:
--  can't adjust the K-value in run-time

I can make necessary changes to load K from a config file if you think using k-way heap is a good idea.

support/src/indirect_intrusive_heap.h
smbillah: yes, there is a  compile-time assertion in the CMakeFiles, so we can get rid of it.

And yes, K = 1 is terrible of course, but it is technically a valid value for K.

ivancich:
Please sign all commits. It's a policy under ceph. You may need to rebase to do so.

PR #14, opened 2016-08-15
Idle client -- turned to PR to make comparisons easier

ivancich:


ivancich:
This code makes a fundamental change to the dmclock algorithm without sufficient justification. Where currently a new client or a client that returns to activity from the idle state will have priority tags that match those of the lowest priority tags by non-idle clients, this code would no longer make that guarantee, although on occasion it would match that. This could be merged, possibly protected by IFDEFS, if a more thorough analysis were done to justify it.

PR #24, opened 2017-02-23
Modify each client's QoS parameter applied time from client_rec creation time to each request's tagging creation time.

bspark8:
The dmclock library updates each client's QoS parameter information when the client related class is created, which is fixed during system online time. it caused the inconvenience of restarting the system when the user want to change the QoS parameter.
Thus I modified each client's QoS parameter applied time from client related class creation time to each request's tagging creation time to support each client's online QoS parameter changes.

### DIFF
--- a/src/dmclock_server.h
+++ b/src/dmclock_server.h
@@ -66,15 +66,15 @@ namespace crimson {
     constexpr uint tag_modulo = 1000000;
 
     struct ClientInfo {
-      const double reservation;  // minimum
-      const double weight;       // proportional
-      const double limit;        // maximum
+      double reservation;  // minimum
+      double weight;       // proportional
+      double limit;        // maximum
 
       // multiplicative inverses of above, which we use in calculations
       // and don't want to recalculate repeatedly
-      const double reservation_inv;
-      const double weight_inv;
-      const double limit_inv;
+      double reservation_inv;
+      double weight_inv;
+      double limit_inv;
 
       // order parameters -- min, "normal", max
       ClientInfo(double _reservation, double _weight, double _limit) :
@@ -229,9 +229,10 @@ namespace crimson {
     }; // class RequestTag
 
 
-    // C is client identifier type, R is request type, B is heap
-    // branching factor
-    template<typename C, typename R, uint B>
+    // C is client identifier type, R is request type,
+    // U1 determines whether to use client information function dynamically,
+    // B is heap branching factor
+    template<typename C, typename R, bool U1, uint B>
     class PriorityQueueBase {
       // we don't want to include gtest.h just for FRIEND_TEST
       friend class dmclock_server_client_idle_erase_Test;
@@ -285,7 +286,7 @@ namespace crimson {
       // ClientRec could be "protected" with no issue. [See comments
       // associated with function submit_top_request.]
       class ClientRec {
-	friend PriorityQueueBase<C,R,B>;
+	friend PriorityQueueBase<C,R,U1,B>;
 
 	C                     client;
 	RequestTag            prev_tag;
@@ -414,7 +415,7 @@ namespace crimson {
 
 	friend std::ostream&
 	operator<<(std::ostream& out,
-		   const typename PriorityQueueBase<C,R,B>::ClientRec& e) {
+		   const typename PriorityQueueBase<C,R,U1,B>::ClientRec& e) {
 	  out << "{ ClientRec::" <<
 	    " client:" << e.client <<
 	    " prev_tag:" << e.prev_tag <<
@@ -543,6 +544,24 @@ namespace crimson {
       }
 
 
+      void update_client_info(const C& client_id) {
+	DataGuard g(data_mtx);
+	auto client_it = client_map.find(client_id);
+	if (client_map.end() != client_it) {
+	  ClientRec& client = (*client_it->second);
+	  client.info = client_info_f(client_id);
+	}
+      }
+
+
+      void update_client_infos() {
+	DataGuard g(data_mtx);
+	for (auto i : client_map) {
+	  i.second->info = client_info_f(i.second->client);
+	}
+      }
+
+
       friend std::ostream& operator<<(std::ostream& out,
 				      const PriorityQueueBase& q) {
 	std::lock_guard<decltype(q.data_mtx)> guard(q.data_mtx);
@@ -651,7 +670,8 @@ namespace crimson {
 	}
       };
 
-      ClientInfoFunc       client_info_f;
+      ClientInfoFunc        client_info_f;
+      static constexpr bool is_dynamic_cli_info_f = U1;
 
       mutable std::mutex data_mtx;
       using DataGuard = std::lock_guard<decltype(data_mtx)>;
@@ -744,6 +764,14 @@ namespace crimson {
       }
 
 
+      inline const ClientInfo get_cli_info(ClientRec& client) const {
+	if (is_dynamic_cli_info_f) {
+	  client.info = client_info_f(client.client);
+	}
+	return client.info;
+      }
+
+
       // data_mtx must be held by caller
       void do_add_request(RequestRef&& request,
 			  const C& client_id,
@@ -831,7 +859,7 @@ namespace crimson {
 
 	if (!client.has_request()) {
 	  tag = RequestTag(client.get_req_tag(),
-			   client.info,
+			   get_cli_info(client),
 			   req_params,
 			   time,
 			   cost);
@@ -840,7 +868,7 @@ namespace crimson {
 	  client.update_req_tag(tag, tick);
 	}
 #else
-	RequestTag tag(client.get_req_tag(), client.info, req_params, time, cost);
+	RequestTag tag(client.get_req_tag(), get_cli_info(client), req_params, time, cost);
 	// copy tag to previous tag for client
 	client.update_req_tag(tag, tick);
 #endif
@@ -890,7 +918,7 @@ namespace crimson {
 #ifndef DO_NOT_DELAY_TAG_CALC
 	if (top.has_request()) {
 	  ClientReq& next_first = top.next_request();
-	  next_first.tag = RequestTag(tag, top.info,
+	  next_first.tag = RequestTag(tag, get_cli_info(top),
 				      top.cur_delta, top.cur_rho,
 				      next_first.tag.arrival);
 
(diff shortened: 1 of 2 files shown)


ivancich:
I can see why this would be helpful. We don't know if the call to client_info_f might be expensive, but if it is, it could be addressed by a later modification to call to see if the parameters have changed and only calling client_info_f if it's aware of changes.

Why is the information stored in the RequestTag (reservation_inv) even if it's also stored in the ClientRec (popped_resv_inv)? Could the ClientRec's value be a previous value if there was a change?

bspark8:
First, yes, you're right. At first, I considered the notification method as you said.
However, in some cases, I realized that the dynamic calling method is somewhat simpler and lighter than we thought.

For example, I will use the dmclock library in Ceph.
In case of per-pool dmclock parameters, dmclock parameters will be stored at OSD map which includes the each pool information. when certain pool's QoS parameters are changed, OSD map will be updated with the new epoch and the each OSD which has older OSD map will try to take new OSD map. And the OSD will try to call ClientRec update procedure through dmclock library. But right then, maybe all existing ClientRecs will be updated because the OSD can't get certain pool's ID from OSD map update event.
To handle above case, quite amount of Ceph codes will be modified.
Inserting ClientRec update procedure to all OSD map updated cases also need quite a bit of Ceph code modification.

For generalization, it would be nice to provide both dynamic calling & notification method in dmclock library.

Second is about reservation_inv.
That reason is simple. When we use the dynamic calling method, as you said, the cost of client_info_f can be expensive. Thus to save a little cost, I'd like to avoid the expensive floating point division.
For example, If we remove reservation_inv from the RequestTag, the code will be changed from "top.set_popped_resv_inv(first.tag.reservation_inv);" to "top.set_popped_resv_inv(0.0 == first.tag.reservation ? 0.0 : 1.0 / first.tag.reservation)".
But, if that does not help performance, it would be better to remove reservation_inv from the RequestTag.

bspark8:
I have added the work you recommended.

First, I added option DO_DYNAMIC_CLI_INFO_F.
If the DO_DYNAMIC_CLI_INFO_F is explicitly defined, the dmclock library the user can call client_info_f function dynamically when creating request's tagging.
It will help the user to use dmclock library with flexibility.
The default option is the original (previous) mode.

Second, I added update_client_info API in order to support the notification method.
This allow user to update the certain ClientRec's ClientInfo at the original (previous) mode.

src/dmclock_server.h
```diff
@@ -1084,6 +1114,24 @@ namespace crimson {
...
 	delete_from_heap(client, ready_heap);
       }
+
+
+      // data_mtx must be held by caller
+      void do_update_client_info(const C& client_id) {
+	auto client_it = client_map.find(client_id);
+	if (client_map.end() != client_it) {
+	  ClientRec& client = (*client_it->second);
+	  client.info = client_info_f(client_id);
```
ivancich: Has this code been tested? ClientInfo declares all its variables as const, and it has no assignment operator. If this is to work, I think we'll have to change ClientInfo so the variables are no longer const and there is an assignment operator.

src/dmclock_server.h
```diff
@@ -1084,6 +1114,24 @@ namespace crimson {
...
+	}
+      }
+
+
+      // data_mtx must be held by caller
+      void do_update_client_infos() {
+	for (auto i : client_map) {
+	  i.second->info = client_info_f(i.second->client);
```
ivancich: I think this will have the same issues outlined above with do_update_client_info.

src/dmclock_server.h
```diff
@@ -1292,6 +1340,18 @@ namespace crimson {
       } // pull_request
 
 
+      void update_client_info(const C& client_id) {
```
ivancich: This function is repeated for both PullPriorityQueue and PushPriorityQueue. Shouldn't this be a public function in PriorityQueueBase?

src/dmclock_server.h
```diff
@@ -1292,6 +1340,18 @@ namespace crimson {
...
+	typename super::DataGuard g(this->data_mtx);
+	super::do_update_client_info(client_id);
+      }
+
+
+      void update_client_info() {
```
ivancich: This function is repeated for both PullPriorityQueue and PushPriorityQueue. Shouldn't this be a public function in PriorityQueueBase?

src/dmclock_server.h
```diff
@@ -1453,6 +1513,18 @@ namespace crimson {
 #endif
       }
 
+
+      void update_client_info(const C& client_id) {
```
ivancich: This function is repeated for both PullPriorityQueue and PushPriorityQueue. Shouldn't this be a public function in PriorityQueueBase?

src/dmclock_server.h
```diff
@@ -1453,6 +1513,18 @@ namespace crimson {
...
+	typename super::DataGuard g(this->data_mtx);
+	super::do_update_client_info(client_id);
+      }
+
+
+      void update_client_info() {
```
ivancich: This function is repeated for both PullPriorityQueue and PushPriorityQueue. Shouldn't this be a public function in PriorityQueueBase?

src/dmclock_server.h
```diff
@@ -821,14 +831,23 @@ namespace crimson {
 	RequestTag tag(0, 0, 0, time);
 
 	if (!client.has_request()) {
-	  tag = RequestTag(client.get_req_tag(), client.info,
-			   req_params, time, cost);
+#ifndef DO_DYNAMIC_CLI_INFO_F
```
ivancich: Looking at how to test this thoroughly along with the existing DO_NOT_DELAY_TAG_CALC, since they interact, you'd have to build the code in four different ways with each combination of the settings. For that reason I'm wondering whether if DO_DYNAMIC_CLI_INFO_F would be better as a boolean template parameter to the PriorityQueue classes. Would you please explore that? You don't have to do anything with DO_NOT_DELAY_TAG_CALC, as if you can successfully change DO_DYNAMIC_CLI_INFO_F to a template parameter, I might then make an effort to do the same with DO_NOT_DELAY_TAG_CALC.

bspark8:
Thank you for your feedback!
I completed rebase & code rearrangement and checked a full unit test pass for the four cases you mentioned.
In addition, I added dmclock_server_pull.update_client_info unit test for update_client_info.
Also, for dynamic_cli_info_f, I added dmclock_server_pull.dynamic_cli_info_f unit test.

src/dmclock_server.h
bspark8: As you recommended, I removed DO_DYNAMIC_CLI_INFO_F and used the boolean template parameter to the PriorityQueueBase class.

src/dmclock_server.h
bspark8: Great catch! Thank you.

src/dmclock_server.h
bspark8: I removed repeated update_client_info(s) in both PullPriorityQueue and PushPriorityQueue.

src/dmclock_server.h
```diff
@@ -744,6 +764,14 @@ namespace crimson {
       }
 
 
+      inline const ClientInfo get_cli_info(ClientRec& client) const {
+	if (dynamic_cli_info_f) {
```
ivancich: Since this is the only use of the variable `dynamic_cli_info_f`, and I assume you like a well-named variable here rather than `U1`, why not put this in the line before
`constexpr bool dynamic_cli_info_f = U1;`
and then use `dynamic_cli_info_f` in your conditional?

src/dmclock_server.h
```diff
@@ -696,6 +715,7 @@ namespace crimson {
       bool             allow_limit_break;
 
       std::atomic_bool finishing;
+      bool             dynamic_cli_info_f = U1;
```
ivancich: We can remove this variable and still get the use of a well-named variable by using a `constexpr` in your `get_cli_info function` below.

test/test_dmclock_server.cc
```diff
@@ -639,6 +639,179 @@ namespace crimson {
...
+
+      ClientId client1 = 17;
+      ClientId client2 = 98;
+
+      dmc::ClientInfo info1(0.0, 1.0, 0.0);
+      dmc::ClientInfo info2(0.0, 2.0, 0.0);
+
+      QueueRef pq;
+
+      auto client_info_f = [&] (ClientId c) -> dmc::ClientInfo {
```
ivancich: To test the dynamic nature of the client_info function, wouldn't we need this function to return a different ClientInfo object for the same client at different points in the run? In other words at some point you might want to change the values of, say, info1, and see if that affects later pulls from the queue.

test/test_dmclock_server.cc
```diff
@@ -639,6 +639,179 @@ namespace crimson {
     } // dmclock_server_pull.pull_reservation
 
 
+    TEST(dmclock_server_pull, update_client_info) {
```
ivancich: This test has failures when `DO_NOT_DELAY_TAG_CALC` is defined.

test/test_dmclock_server.cc
```diff
@@ -639,6 +639,179 @@ namespace crimson {
...
+      }
+
+      EXPECT_EQ(4, c1_count) <<
+	"after: one-third of request should have come from first client";
+      EXPECT_EQ(2, c2_count) <<
+	"after: two-thirds of request should have come from second client";
+    }
+
+
+    TEST(dmclock_server_pull, dynamic_cli_info_f) {
```
ivancich: This test has failures when `DO_NOT_DELAY_TAG_CALC` is defined.

bspark8:
I have left messages and completed the changes requested. Thank you!

src/dmclock_server.h
bspark8: I had originally intended to use well-named variable to support complie time constant, as you said.
The reason I used bool variable was to change the use of dynamic_cli_info_f via the configuration change even after the queue was created.
However, it seems that it is not necessary now.

test/test_dmclock_server.cc
bspark8: Sorry for the lack of check. I was fixed the failure situation.

test/test_dmclock_server.cc
bspark8: Thank you and to test the dynamic nature of the client_info function, I modified the function client_info_f according to your opinion.

ivancich:
Thank you!

PR #25, opened 2017-05-26
dmclock: Don't dump core when using EXPECT_DEATH_IF_SUPPORTED

badone:
Fixes: http://tracker.ceph.com/issues/19979

Signed-off-by: Brad Hubbard <bhubbard@redhat.com>

### DIFF
--- a/test/test_dmclock_server.cc
+++ b/test/test_dmclock_server.cc
@@ -16,6 +16,46 @@
 #include "dmclock_util.h"
 #include "gtest/gtest.h"
 
+#include <dmtest-config.h>
+#ifdef HAVE_SYS_PRCTL_H
+#include <iostream>
+#include <sys/prctl.h>
+#include <errno.h>
+
+struct PrCtl {
+  int saved_state = -1;
+  int set_dumpable(int new_state) {
+    int r = prctl(PR_SET_DUMPABLE, new_state);
+    if (r) {
+      r = -errno;
+      std::cerr << "warning: unable to " << (new_state ? "set" : "unset")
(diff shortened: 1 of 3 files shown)


badone:
@ivancich you stated you didn't want to use that and it was I that suggested it originally?

ivancich:
Sorry, I guess I wasn't clear. What I meant is that we could not include that specific file since it exists within ceph and not the dmclock library. I have no problem with the implementation, though. Sorry for the confusion.

badone:
No problem, will take another run at this shortly.

badone:
@ivancich like this?

test/test_dmclock_server.cc
```diff
@@ -16,6 +16,46 @@
 #include "dmclock_util.h"
 #include "gtest/gtest.h"
 
+#include <dmtest-config.h>
+#ifdef HAVE_SYS_PRCTL_H
+#include <iostream>
+#include <sys/prctl.h>
+#include "common/errno.h"
```
ivancich: common/errno.h is specific to the ceph build tree. dmclock needs to build independently (i.e., "git clone git@github.com:ceph/dmclock.git" and then build). I tried removing this #include and it seemed to build and run tests and the simulation. The build system had /usr/include/sys/prctl.h and according to CMakeCache.txt "HAVE_SYS_PRCTL_H:INTERNAL=1". So is that #include necessary?

ivancich: I'm currently do a full ceph build and test.

badone: Sorry, should have changed this to <errno.h>

PR #26, opened 2017-06-14
set compiler flags via CMAKE_BUILD_TYPE to fix the error of compilation

joelin:
when i try to compile dmclock at my computer, i got an error. try to set compiler flags via CMAKE_BUILD_TYPE to fix the error of compilation
`
In file included from /usr/include/c++/4.8.2/atomic:38:0,
                 from /root/dmclock/test/test_test_client.cc:7:
/usr/include/c++/4.8.2/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support is currently experimental, and must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
 #error This file requires compiler and library support for the \
  ^
In file included from /usr/include/c++/4.8.2/atomic:41:0,
                 from /root/dmclock/test/test_test_client.cc:7:
`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,12 @@ cmake_minimum_required(VERSION 2.8.11)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake/modules")
 
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wno-write-strings -Wall -pthread")
+
+if(DO_NOT_DELAY_TAG_CALC)
+  SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DDO_NOT_DELAY_TAG_CALC")
+endif()
+
 if (NOT(TARGET gtest AND TARGET gtest_main))
   if(NOT(GTEST_FOUND))
     find_package(GTest REQUIRED)


CMakeLists.txt
```diff
@@ -2,6 +2,12 @@ cmake_minimum_required(VERSION 2.8.11)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake/modules")
 
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wno-write-strings -Wall -pthread")
+
+if(DO_NOT_DELAY_TAG_CALC)
```
ivancich: Does this conditional need to be included? The conditional appears in src/CMakeLists.txt and the only references to DO_NOT_DELAY_TAG_CALC are under the src subdirectory.

Or in other words, does something break if you were to remove the entire conditional from here?

ivancich:
Which version of g++ are you using? I'm using 6.3.1 and apparently with gcc version 6, the default standard became -std=gnu++14. That would explain why I'm getting different results. Nonetheless, we should specify the version. Thanks!

ivancich:
Nevermind, from your error message I see that you're using 4.8.2.

joelin:
the env is :

[root@test1 ganesha]# g++ --version
g++ (GCC) 4.8.5 20150623 (Red Hat 4.8.5-11)
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

[root@test1 ganesha]# lsb_release -a
LSB Version:	:core-4.1-amd64:core-4.1-noarch
Distributor ID:	CentOS
Description:	CentOS Linux release 7.3.1611 (Core)
Release:	7.3.1611
Codename:	Core

joelin:
does it means i need update the version  to  6.X ?

ivancich:
> does it means i need update the version to 6.X ?

No, I was just curious as to why it worked in my environment but not in yours. I'm happy to add your change to include "-std=c++11". My only question is whether the conditional for DO_NOT_DELAY_TAG_CALC is necessary.

joelin:
i am not sure.. fixed the problem  just reference from the source code `src/CMakeLists.txt`  in my environment.

```
set(local_flags "-std=c++11 -Wno-write-strings -Wall -pthread")

//reference from this 
if(DO_NOT_DELAY_TAG_CALC)
  set(local_flags "${local_flags} -DDO_NOT_DELAY_TAG_CALC")
endif()

set(dmc_srcs dmclock_util.cc ../support/src/run_every.cc)

//doesn't work in my environment
set_source_files_properties(${dmc_srcs}
  PROPERTIES
  COMPILE_FLAGS "${local_flags}"
  )
```

joelin:
i think cmake_<lang>_flags is better than manual set the compile flags

ivancich:
I understand you just copied this from the other file:

if(DO_NOT_DELAY_TAG_CALC)
  set(local_flags "${local_flags} -DDO_NOT_DELAY_TAG_CALC")
endif()

But why did you copy it? What purpose does it serve? You need to either explain why it's necessary or remove it.

joelin:
i want to keep the all compile flags the same, via `cmake_<lang>_flags`.
it served An optimization/fix to the published algorithm has been added and is on by default
To disable this optimization/fix run cmake with:
> -DDO_NOT_DELAY_TAG_CALC=yes

ivancich:
Since DO_NOT_DELAY_TAG_CALC is only used in the src subdirectory, my intention is that it should only appear in the cmake file(s) in the src subdirectory.

I made an alternate pull request that incorporates some of your change and does other cmake clean-up. https://github.com/ceph/dmclock/pull/27 Would it be acceptable to you?

joelin:
I accept！

ivancich:
Awesome! Thank you for all your efforts!

PR #29, opened 2017-07-12
cmake: build gtest if not found

tchaikov:
debian stopped shipping googletest binaries, but it offers googletest in
the source form. so we can use it instead.

Signed-off-by: Kefu Chai <kchai@redhat.com>

PR #30, opened 2017-07-12
enforce move semantics of request types

tchaikov:
this requires the request type support move semantics, but the upside is
that the request type will not be copied under any circumstances.

Signed-off-by: Kefu Chai <kchai@redhat.com>

### DIFF
--- a/sim/src/sim_client.h
+++ b/sim/src/sim_client.h
@@ -89,7 +89,7 @@ namespace crimson {
 
       using SubmitFunc =
 	std::function<void(const ServerId&,
-			   const TestRequest&,
+			   TestRequest&&,
 			   const ClientId&,
 			   const ReqPm&)>;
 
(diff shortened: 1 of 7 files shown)


ivancich:
The build command:

> make dmclock-sims

fails. I think you need to change the sim to use move semantics as well.

tchaikov:
@ivancich fixed and repushed.

ivancich:
@tchaikov What do you think about in the future supporting both move semantics and const-ref-copying? Thinking about the general case, outside of ceph, I wonder if some users of the library would prefer their requests not be clobbered. Maybe only allow copy semantics *if* compiled with a certain macro defined or if instantiated with a template bool parameter set.

I'm not asking you to do this. I'm just curious as to your thoughts.

Thanks!

tchaikov:
@ivancich i am thinking about generalizing this library by using traits to selectively use copy semantics for copyable requests, and use move semantics as long as the request type supports move semantics. as we can tell if a type supports move semantics by using `std::is_move_constructible` and `std::is_move_assignable`.

ivancich:
@tchaikov I recall you were working on something else and wanted to delay bringing this into ceph master until that was one. Can you remind me what that was and of its status?

tchaikov:
@ivancich see https://github.com/ceph/dmclock/pull/29#issuecomment-315406794 and i am querying @myoungwon on his progress on https://github.com/ceph/ceph/pull/16030

PR #39, opened 2017-09-22
Made delta/rho calculations more flexible via tracker plug-in; make default a borrowing tracker

ivancich:
Allow the calculations of rho and delta to be handled by a "tracker"
specified via template parameter (i.e., by static polymorphism). The
tracker follows a simple interface constisting of three functions and
one static function.

Add a BorrowingTracker. Make it the default tracker. Add unit tests
for it. Maintain the original calculations as the OrigTracker.

### DIFF
--- a/sim/src/test_dmclock.h
+++ b/sim/src/test_dmclock.h
@@ -29,13 +29,14 @@ namespace crimson {
     };
 
     using DmcQueue = dmc::PushPriorityQueue<ClientId,sim::TestRequest>;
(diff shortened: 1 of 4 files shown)


TaewoongKim:
@ivancich I think it would be better to do that too. BorrowingTracker included a part of a concept that was intended by #23. If there is some other thing that I wanna to include to BorrowingTrack I will make another PR later. Thank you for asking my opinion.

PR #42, opened 2017-10-18
cmake,travis: add travis CI on GNU/Linux

tchaikov:


tchaikov:
@ivancich Hey Eric, i think it'd be great if we can have a CI for dmclock also. but i don't have the admin privilege to enable it at https://travis-ci.org/ceph/dmclock. if you think it's good to have, could you help to do it?

ivancich:
@tchaikov I've thought about that (or something similar). Currently PRs have a check to see whether someone submits code that modifies a git submodule. I'd like a similar check to provide a warning if they modified code in a git subtree. Sometimes the divergence is good, but most often we'd like the change in the original library.

But I'm guessing you're referring to ci when changes are made in the ceph/dmclock repo. That too would be useful. And I'm happy to work on it.

I'm unfamiliar with travis-ci, so I'll need to do some reading....

tchaikov:
yes, i want to address the CI of ceph/dmclock itself.

agreed, that'd a pain if new developers keep changing the code in subtree via ceph/ceph. probably we can switch to submodule ;-)

tchaikov:
@ivancich ping?

ivancich:
@tchaikov My apologies! Working on it now.

ivancich:
@tchaikov Wasn't sure about the next step. When I tried to link to travis-ci it wanted many permissions to read (and sometimes write) some aspects of my github account, and that made me uncomfortable. I thought perhaps that I should create another github account to act as an integrator between ceph/dmclock and travis-ci.

But since you seem to already understand travis-ci I just gave you admin privileges on ceph/dmclock, so you could determine the best course of action. Note: If I add travis-ci as an integration via the github interface it wants three pieces of information -- User, Token, and Domain.

Thank you for your help and sorry for my delay.

tchaikov:
@ivancich i activated the travis-ci for this repo using my account. so the master branch will built when it is updated. i just triggered a built at https://travis-ci.org/ceph/dmclock/builds/292929175. 

and i believe the new PRs will be checked also.

ivancich:
That's cool! I like that it builds with both gcc and clang. When there's a build for a PR will there be CI info on github's page for the PR (like what we have with ceph-ci)? Or will people have to know to check the travis-ci page?

Thanks for doing this!

tchaikov:
@ivancich the test against PRs would look like https://github.com/sunpinyin/sunpinyin/pull/72. search "All checks have passed", and by clicking on "Show all checks" will bring the user to travis CI's build page. it works just like our own "jenkins make check" run. and i believe both of them are using the same github API. so the don't need to check the travis-ci page.

tchaikov:
we can also add a badge in our readme showing the build status just like https://github.com/sunpinyin/sunpinyin and https://github.com/ceph/ceph-container, if you think it's a good idea =)

PR #43, opened 2017-10-27
Add missing anticipation_timeout argument to PullPriorityQueue constructor

TaewoongKim:
I missed adding anticipation_timeout argument for a constructor at the previous PR.

If this PR is merged I will make a ceph PR that makes anticipation timeout be configurable for ceph.

The code will be https://github.com/TaewoongKim/ceph/commit/8f17269f534dfbf7acebd7fd5ad251df4592b2c8

### DIFF
--- a/src/dmclock_server.h
+++ b/src/dmclock_server.h
@@ -1206,12 +1206,14 @@ namespace crimson {
 
       // pull convenience constructor
       PullPriorityQueue(typename super::ClientInfoFunc _client_info_f,
-			bool _allow_limit_break = false) :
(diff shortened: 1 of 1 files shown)


ivancich:
@TaewoongKim Is it OK if we don't pull this into ceph immediately? There's a bit of work to check everything and we just did a pull not that long ago, so when possible I like to accumulate commits before the pull. But please let me know if you have any timing constraints.

TaewoongKim:
@ivancich OK, let me know if it's ready.

PR #45, opened 2017-11-08
Reduce the size of enum class PhaseType

TaewoongKim:
I think it's too large using default enum class type for the PhaseType.
Rather than using 4 bytes, It would be better to use only 1 byte.

Furthermore, the current ceph's denc.h implementation doesn't support 4 or 2 bytes underlying typed enum. (I don't know it's intended or not).

Anyway, I think decreasing the size would be good.

### DIFF
--- a/src/dmclock_recs.h
+++ b/src/dmclock_recs.h
@@ -17,7 +17,7 @@ namespace crimson {
   namespace dmclock {
     using Counter = uint64_t;
 
-    enum class PhaseType { reservation, priority };
+    enum class PhaseType : uint8_t { reservation, priority };
 
     inline std::ostream& operator<<(std::ostream& out, const PhaseType& phase) {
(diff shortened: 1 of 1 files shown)


tchaikov:
@TaewoongKim i don't think it's intended. my guess is that we don't have a user for `{u,}int16_t` or `{u,}int32_t` 's denc yet. please feel free to add them.

TaewoongKim:
@tchaikov I think it's intended. There are comments about that.
https://github.com/ceph/ceph/pull/18701/commits/3fa4e8109e10c3b60b73d333ca4932775ab85e63#diff-74763254ed0f629b4f02d9e89bafb31fR311

tchaikov:
oh, what i mean is, it's not intended to exclude **{u,}int16_t** and **{u,}int32_t** from the supported types.

ivancich:
Good catch!

PR #46, opened 2018-01-25
Allow the dmclock library to handle the "cost" of an operation/request

ivancich:
Allow the dmclock library to handle the "cost" of an operation/request. This is in contrast to just measuring ops. If the cost is set at the value 1, then it's equivalent to the old code.

The cost is calculated on the server side and affects the tag values for the request. Furthermore, the cost is then sent back to the client, so the ServiceTracker can use the cost to properly calculate delta and rho values.

We now allow the delta and rho values sent with a request to be zero, since the request cost is included on the server-side tag calculation, and the request cost must be positive guaranteeing the advancement of tags.

The OrigTracker has now been updated so as not to add one when computing delta and rho.

The unit tests have been updated to use request costs. And the simulator has been updated to use cost, and the cost of a client group's requests can be configured.

The simulation has been changed to user the OrigTracker as the default tracker.

### DIFF
--- a/sim/src/config.h
+++ b/sim/src/config.h
@@ -23,6 +23,8 @@
 
 #include "ConfUtils.h"
 
+#include "sim_recs.h"
+
 
 namespace crimson {
   namespace qos_simulation {
@@ -37,6 +39,7 @@ namespace crimson {
       double client_reservation;
       double client_limit;
       double client_weight;
+      Cost client_req_cost;
 
       cli_group_t(uint _client_count = 100,
 		  uint _client_wait = 0,
@@ -46,7 +49,8 @@ namespace crimson {
 		  uint _client_outstanding_ops = 100,
 		  double _client_reservation = 20.0,
 		  double _client_limit = 60.0,
-		  double _client_weight = 1.0) :
+		  double _client_weight = 1.0,
+		  Cost _client_req_cost = 1u) :
 	client_count(_client_count),
 	client_wait(std::chrono::seconds(_client_wait)),
 	client_total_ops(_client_total_ops),
@@ -55,7 +59,8 @@ namespace crimson {
 	client_outstanding_ops(_client_outstanding_ops),
 	client_reservation(_client_reservation),
 	client_limit(_client_limit),
-	client_weight(_client_weight)
+	client_weight(_client_weight),
+	client_req_cost(_client_req_cost)
       {
 	// empty
       }
@@ -72,7 +77,8 @@ namespace crimson {
 	  std::fixed << std::setprecision(1) <<
 	  "client_reservation = " << cli_group.client_reservation << "\n" <<
 	  "client_limit = " << cli_group.client_limit << "\n" <<
-	  "client_weight = " << cli_group.client_weight;
+	  "client_weight = " << cli_group.client_weight << "\n" <<
+	  "client_req_cost = " << cli_group.client_req_cost;
 	return out;
       }
     }; // class cli_group_t
--- a/sim/src/test_dmclock.cc
+++ b/sim/src/test_dmclock.cc
@@ -12,6 +12,7 @@
  * COPYING.
  */
 
+#include <type_traits>
 
 #include "dmclock_recs.h"
 #include "dmclock_server.h"
@@ -25,6 +26,13 @@
 
 
 namespace test = crimson::test_dmc;
+
+
+// Note: if this static_assert fails then our two definitions of Cost
+// do not match; change crimson::qos_simulation::Cost to match the
+// definition of crimson::dmclock::Cost.
+static_assert(std::is_same<crimson::qos_simulation::Cost,crimson::dmclock::Cost>::value,
+	      "Please make sure the simulator type crimson::qos_simulation::Cost matches the dmclock type crimson::dmclock::Cost.");
 
 
 void test::dmc_server_accumulate_f(test::DmcAccum& a,
--- a/src/dmclock_client.h
+++ b/src/dmclock_client.h
@@ -45,7 +45,7 @@ namespace crimson {
     public:
 
       OrigTracker(Counter global_delta,
-		 Counter global_rho) :
+		  Counter global_rho) :
 	delta_prev_req(global_delta),
 	rho_prev_req(global_rho),
 	my_delta(0),
@@ -57,8 +57,8 @@ namespace crimson {
       }
 
       inline ReqParams prepare_req(Counter& the_delta, Counter& the_rho) {
-	Counter delta_out = 1 + the_delta - delta_prev_req - my_delta;
-	Counter rho_out = 1 + the_rho - rho_prev_req - my_rho;
+	Counter delta_out = the_delta - delta_prev_req - my_delta;
+	Counter rho_out = the_rho - rho_prev_req - my_rho;
 	delta_prev_req = the_delta;
 	rho_prev_req = the_rho;
 	my_delta = 0;
@@ -68,12 +68,13 @@ namespace crimson {
 
       inline void resp_update(PhaseType phase,
 			      Counter& the_delta,
-			      Counter& the_rho) {
-	++the_delta;
-	++my_delta;
+			      Counter& the_rho,
+			      Cost cost) {
+	the_delta += cost;
+	my_delta += cost;
 	if (phase == PhaseType::reservation) {
-	  ++the_rho;
-	  ++my_rho;
+	  the_rho += cost;
+	  my_rho += cost;
 	}
       }
 
@@ -139,10 +140,11 @@ namespace crimson {
 
       inline void resp_update(PhaseType phase,
 			      Counter& the_delta,
-			      Counter& the_rho) {
-	++the_delta;
+			      Counter& the_rho,
+			      Counter cost) {
+	the_delta += cost;
 	if (phase == PhaseType::reservation) {
-	  ++the_rho;
+	  the_rho += cost;
 	}
       }
 
@@ -152,9 +154,13 @@ namespace crimson {
     }; // struct BorrowingTracker
 
 
-    // S is server identifier type
-    // T is the server info class that adheres to ServerTrackerIfc interface
-    template<typename S, typename T = BorrowingTracker>
+    /*
+     * S is server identifier type
+     *
+     * T is the server info class that adheres to ServerTrackerIfc
+     * interface
+     */
+    template<typename S, typename T = OrigTracker>
     class ServiceTracker {
       // we don't want to include gtest.h just for FRIEND_TEST
       friend class dmclock_client_server_erase_Test;
@@ -210,9 +216,11 @@ namespace crimson {
 
 
       /*
-       * Incorporates the RespParams received into the various counter.
+       * Incorporates the response data received into the counters.
        */
-      void track_resp(const S& server_id, const PhaseType& phase) {
+      void track_resp(const S& server_id,
+		      const PhaseType& phase,
+		      Counter request_cost = 1u) {
 	DataGuard g(data_mtx);
 
 	auto it = server_map.find(server_id);
@@ -224,7 +232,7 @@ namespace crimson {
 				      T::create(delta_counter, rho_counter));
 	  it = i.first;
 	}
-	it->second.resp_update(phase, delta_counter, rho_counter);
+	it->second.resp_update(phase, delta_counter, rho_counter, request_cost);
       }
 
       /*
--- a/src/dmclock_recs.h
+++ b/src/dmclock_recs.h
@@ -24,6 +24,12 @@ namespace crimson {
   namespace dmclock {
     using Counter = uint64_t;
 
+    // we're abstracting cost to its own type to better allow for
+    // future changes; we're assuming that Cost is relatively small
+    // and that it would be more efficient to pass-by-value than
+    // by-reference.
+    using Cost = uint32_t;
+
     enum class PhaseType : uint8_t { reservation, priority };
 
     inline std::ostream& operator<<(std::ostream& out, const PhaseType& phase) {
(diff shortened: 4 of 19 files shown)


ivancich:
@bspark8 @TaewoongKim Please let me know your thoughts on this PR. Thanks!

src/dmclock_server.h
```diff
@@ -199,15 +208,14 @@ namespace crimson {
...
+			     uint64_t cost) {
 	if (0.0 == increment) {
 	  return extreme_is_high ? max_tag : min_tag;
 	} else {
-	  if (0 != dist_req_val) {
-	    increment *= dist_req_val;
-	  }
-	  return std::max(time, prev + increment);
+	  double tag_increment = increment * (dist_req_val + cost);
+	  return std::max(time, prev + tag_increment);
```
TaewoongKim: I'm not digging the code in depth yet. Sorry about that.
I just looked the code roughly and I have a question about the manner dealing the cost.
In the code, the cost is usually added to rho or delta value. However, I think cost is somewhat needed to be multiplied. Because cost is a relative value. So it should represent how much times consume the resource than the unit request(e.g. 8KB write request.)
In the dmclock paper, author said "So, for tagging purposes, a single request of IO size S is treated as equivalent to: (1+S/(Tm×Bpeak)) IO requests"
I understood that sentence as (1+S/(Tm×Bpeak)) is something like a cost and it should be multiplied with tag incremental value to deal the IO request as multiple IO requests.

tchaikov: yeah, i feel the same. cost is quite the same as the number of unit requests. and both \rho_{i} and \delta_{i} denote the number of served requests. if this certain request is treated a multiple unit requests, we should certainly use something like
ql_b2f61a2495fb7623243337d1135b7ea0_l3

ivancich: Thank you both for your detailed look into this. This is the type of discussion I was hoping we'd have!

So cost is calculated on the server side by a formula as discussed in the [ceph-devel discussion](https://marc.info/?l=ceph-devel&m=151362386931501&w=2). So yes, on a hard disk, it would have to incorporate seek times and read/write times.

One thing to keep in mind is that cost, rho, and delta are all in the same units. On the client side, when the response comes back, delta and possibly rho (if the request was handled during the reservation phase) are incremented by the cost value. See dmclock_client.h:dmclock::SimpleTracker::resp_update.

So delta (or rho) represents costs of services already received, and cost represents cost of services being scheduled to be received. Because they are in the same unit, they should not be multiplied but instead be added. It's not clear to me what cost_squared units would represent.

TaewoongKim: Cost doesn't include unit concept. It's relative value like weight or ratio value that don't have any unit info.
Therefore, we need another unit concept that represents an abstract amount of resources that would be needed to process a unit request(e.g 8KB write) so that a cost can be multiplied with this unit and the result can have a meaning in the real world. This means a cost for a specific request will be represented as a ratio to the unit cost(e.g. cost for an 8KB write).
Rho & delta is also a kind of ratio value and these don't have a unit also.
So, we can multiply these values with each other multiple times.

yzhan298: When I reproduce the experiment from the mclock paper, I find the behavior of the dmclock with cost integrated is not the same as expected (when `cost`=1). The reason for that is due to line 244: `double tag_increment = increment * (dist_req_val + cost)`. My understanding is that cost should replace the `delta` and `rho` (or multiply it). In `dmclock_client.h`, we already added the cost into `delta` and `rho`, so here at line 244, we should multiply `increment` with `cost` directly.

tchaikov: @yzhan298 i think you mean line 224, i.e. https://github.com/ceph/dmclock/pull/46/files#diff-9674a65a2e09d3da2ab1320a086feed7R224 right?

yzhan298: @tchaikov yes, I have updated the comment. Thanks

ivancich:
@yzhan298 This is exciting and I can't wait to spend some time looking over your suggested changes. I'm on vacation, but I think I'll have time this weekend. Thanks for your help on this!!

src/dmclock_server.h
ivancich: I'd like to update everyone with some conversations that @yzhan298 and I had. @yzhan298's suggested fix cannot be correct, because it would mean that rho and delta would have no impact on tag calculation, and that cannot be correct. Once we eliminate delta and rho from tag calculation we no longer have "dmclock" but instead "mclock". The modification did produce results from the simulation more in line with expectations. But that should serve as a reminder for the importance of having a theoretical grounding for changes.

But working through that with @yzhan298 made me question the calculation of delta and rho, and that's done in the client-side trackers. So I re-evaluated that and since we're adding cost in on the server side, we no longer have to insure delta and rho are at least 1. So I modified the OrigTracker to get rid of the "1 +" and @yzhan298 was kind enough to test it out, and it produced the expected results.

In the past we'd migrated from the OrigTracker to the BorrowingTracker. The OrigTracker was written to implement the algorithm in the dmclock paper. It subtracts out responses it gets from the same server in the calculation of delta and rho.

But it seems like the new OrigTracker is the way to go.

ivancich: @TaewoongKim You are absolutely correct that _cost_ is a scalar. That is why when we discussed the addition of cost, we decided that we should re-think how we conceptualize how we think about reservation and limit. Currently they are in the unit of ops per second. But it would likely be more appropriate to think of them as _cost units per second_. So a simple/small operation might be the foundation as one cost unit. Operations that require more data to be written/read would represent multiple cost units; the same would be the case for operations that ultimately turn into multiple sub-operations.

We are even contemplating using _cost units per time unit_ that is not seconds in order to make sure reservations and limits are expressed in positive integers given some issues in sending doubles in messages.

For the initial announcement and follow-up conversation, please see the ceph-devel RFC in:
https://marc.info/?l=ceph-devel&m=151362386931501&w=2

ivancich: @tchaikov Yes, rho and delta are defined to count operations. But in the transition to thinking about costs and operations of *widely* varying costs, we need to include cost in rho and delta, which this PR does. They are measures of work done for a client by *other* servers in a given time interval (time since last request to the same server). And they should include not just # of ops but combined cost of those ops. That is why in the tracker the client increments delta by cost (and rho by cost when done in the reservation phase).

So when we calculate the tag for a new request, we need to incorporate the cost of work done by other servers and the calculated cost of this operation. We can think of it as `cost_recent_past` and `cost_this` and we therefore must add them together. Multiplying them together would result in "units" of _cost_squared_, which is not meaningful.

ivancich: @TaewoongKim The part of the dmclock paper that you quote deals specifically with characteristics of a spinning hard disk. We need something more general as we're not working strictly with hard disks, but SSDs and non-volatile memory. Plus the cost of operations sent to the OSD will vary *widely*. So we need to work with a more general concept of cost, it seems.

src/dmclock_server.h
```diff
@@ -112,41 +112,47 @@ namespace crimson {
...
     struct RequestTag {
-      double reservation;
-      double proportion;
-      double limit;
-      bool   ready; // true when within limit
-      Time   arrival;
+      double   reservation;
+      double   proportion;
+      double   limit;
+      uint64_t cost;
```
cbodley: consider exposing a typedef for `Cost` as part of the interface. this would help disambiguate it from other uint64_t parameters, and calling code would be less likely to rely on its exact representation

src/dmclock_server.h
```diff
@@ -112,41 +112,47 @@ namespace crimson {
...
+		 const uint64_t rho,
 		 const Time time,
-		 const double cost = 0.0,
+		 const uint64_t _cost = 1u,
 		 const double anticipation_timeout = 0.0) :
+	cost(_cost),
 	ready(false),
 	arrival(time)
       {
+	assert(cost > 0.0);
```
cbodley: cost is no longer floating point, so `cost > 0` here? (and again on line 180)

for requests that pass cost=0, it's probably better fail with EINVAL error instead of aborting

src/dmclock_server.h
ivancich: Very nice idea. Latest version does this.

src/dmclock_server.h
ivancich: Fixed 0 literal issues.

I don't know that I want to take on improving error returning results in this PR. But I think it's an excellent idea. At the moment there are a lot of asserts that'd need to be tackled.

src/dmclock_recs.h
```diff
@@ -33,20 +33,20 @@ namespace crimson {
 
     struct ReqParams {
       // count of all replies since last request; MUSTN'T BE 0
-      uint32_t delta;
+      uint64_t delta;
```
tchaikov: i am just curious why we want to use `uint64_t` for delta and rho now? because we could be running into integer overflow with `uint32_t`. as IMO,  `uint32_t` is able to represent a fairly large number already.

ivancich: This is a very good point. I think we had a mention that rho and delta would be u64s on our etherpad when we discussed it. But that seems to large. 4 billion should be more than enough to count the number of replies received from other servers. I'll switch it back.

src/dmclock_server.h
```diff
@@ -72,6 +72,12 @@ namespace crimson {
       std::numeric_limits<double>::lowest();
     constexpr uint tag_modulo = 1000000;
 
+    // we're abstracting cost to its own type to better allow for
+    // future changes; we're assuming that Cost is relatively small
+    // and that it would be more efficient to pass-by-value than
+    // by-reference.
+    using Cost = uint64_t;
```
ivancich: @tchaikov's question regarding whether delta and rho should be uint64_t's raises a similar question w.r.t. Cost. Should Cost also be uint32_t? If we use cost=1 as a base, would there be an op that had a relative cost of more than 4 billion? If not, uint32_t would be sufficient.

@cbodley @tchaikov Your thoughts?

src/dmclock_recs.h
cbodley: i can imagine a strategy that sets cost=bytes in a request, and radosgw will happily serve requests bigger than 4G. but given a Cost=uint32_t, you could easily adapt the strategy to some fraction of the bytes instead

are we doing much math with these values, where there'd be risk of an intermediate multiplication overflowing?

it seems like the benefits of uint32_t would be less overhead in osd messages, and less per-request memory for the dmclock prio queue? those saving could definitely add up

src/dmclock_server.h
```diff
@@ -204,17 +219,16 @@ namespace crimson {
...
+			     const bool extreme_is_high,
+			     const Cost cost) {
 	if (0.0 == increment) {
 	  return extreme_is_high ? max_tag : min_tag;
 	} else {
-	  if (0 != dist_req_val) {
-	    increment *= dist_req_val;
-	  }
-	  return std::max(time, prev + increment);
+	  double tag_increment = increment * (dist_req_val + cost);
```
ivancich: @cbodley This is the key calculation with _cost_. We're adding two `uint32_t`'s and then multiplying by a `double`. If we're concerned about overflow, we could cast one of the addends to a uint64_t to force `uint64_t` addition, as in:
`double tag_increment = increment  * (uint64_t(dist_req_val) + cost);`

src/dmclock_recs.h
ivancich: I can imagine someone trying that strategy as well, although I'd hope people would think in terms of blocks rather than bytes. Assuming 4K blocks a request of 4G would result in a cost of 1M. You'd then need a request of 16T to overflow cost.

So I think we're piecing together a case for uint32_t for cost. @tchaikov -- your thoughts?

ivancich:
@tchaikov I'd like to get this merged. You had indicated doubts about how cost was integrated. Have your doubts been addressed?

src/dmclock_server.h
tchaikov: i think 4G is a very large number in the context of a networking protocol at this moment. we can cap the cost at `std::numeric_limit<uint32>::max()` (which is 4G). if its value is actually greater than it.

tchaikov:
@ivancich yes. and regarding to https://github.com/ceph/dmclock/pull/46#discussion_r187615178 , i don't feel strong either way.

sim/src/config.h
```diff
@@ -37,6 +37,7 @@ namespace crimson {
       double client_reservation;
       double client_limit;
       double client_weight;
+      uint64_t client_req_cost;
```
cbodley: can stuff in sim/ use the Cost typedef?

src/dmclock_client.h
```diff
@@ -68,12 +68,13 @@ namespace crimson {
 
       inline void resp_update(PhaseType phase,
 			      Counter& the_delta,
-			      Counter& the_rho) {
-	++the_delta;
-	++my_delta;
+			      Counter& the_rho,
+			      Counter cost) {
```
cbodley: Counter -> Cost?

src/dmclock_server.h
cbodley: you've convinced me that uint32_t is sufficient here

test/test_test_client.cc
```diff
@@ -45,6 +45,7 @@ TEST(test_client, full_bore_timing) {
   sim::TestResponse resp(0);
   dmc::PhaseType resp_params = dmc::PhaseType::priority;
   test::DmcClient* client;
+  const uint64_t request_cost = 1u;
```
cbodley: uint64_t -> Cost

sim/src/config.h
ivancich: I ended up doing an odd work-around here. The simulator defines its own Cost and then I assert upon start-up that the sizeofs of the two are the same. I played a bit with SFINAE but couldn't get it to work and didn't want to spend more time on it. Your thoughts?

sim/src/test_dmclock.cc
```diff
@@ -45,3 +46,13 @@ void test::dmc_client_accumulate_f(test::DmcAccum& a,
     ++a.proportion_count;
   }
 }
+
+
+// Note: this would be better if we could use std::is_same and SFINAE
```
ivancich: @cbodley Here's where I played with SFINAE....

sim/src/test_dmclock.cc
```diff
@@ -27,6 +28,13 @@
 namespace test = crimson::test_dmc;
 
 
+// Note: if this static_assert fails then our two definitions of Cost
+// do not match; change crimson::qos_simulation::Cost to match the
+// definition of crimson::dmclock::Cost.
+static_assert(std::is_same<crimson::qos_simulation::Cost,crimson::dmclock::Cost>::value,
```
tchaikov: nit, could use `std::is_same_v` next time.

ivancich: Thanks; I did not know about it.

PR #50, opened 2018-03-23
add option to reject requests to clients over limit

cbodley:
replaces `bool allow_limit_break` with an `enum AtLimit` to enable a third option, whereby calls to `add_request()` will fail with `EAGAIN` if the client is at/above its limit

add a tunable threshold for `AtLimit::Reject` such that requests are only rejected once the limit is over this threshold. this is intended to help balance the queue depth with the desire to enforce an upper bound on the time spent waiting in the queue (and the memory use associated with queued requests)

for example, setting this threshold to 1 second for a client with limit=10 would allow that client to enqueue 10 requests before any are rejected. with a threshold of 0, all but the first would be rejected and require the client to resubmit

to simplify queue construction, a variant type `AtLimitParam` allows construction with either an explicit `AtLimit` enum value, or with a `RejectThreshold` value that implies `AtLimit::Reject`

### DIFF
--- a/src/dmclock_server.h
+++ b/src/dmclock_server.h
@@ -72,7 +72,8 @@ namespace crimson {
       // are met and below their limits
       Allow,
       // if an incoming request would exceed its limit, add_request() will
-      // reject it with EAGAIN instead of adding it to the queue
+      // reject it with EAGAIN instead of adding it to the queue. cannot be used
+      // with DelayedTagCalc, because add_request() needs an accurate tag
       Reject,
     };
 
@@ -797,6 +798,8 @@ namespace crimson {
       {
 	assert(_erase_age >= _idle_age);
 	assert(_check_time < _idle_age);
+	// AtLimit::Reject depends on ImmediateTagCalc
+	assert(at_limit != AtLimit::Reject || !IsDelayed);
 	cleaning_job =
 	  std::unique_ptr<RunEvery>(
 	    new RunEvery(check_time,
--- a/src/dmclock_server.h
+++ b/src/dmclock_server.h
@@ -846,11 +846,11 @@ namespace crimson {
       }
 
       // data_mtx must be held by caller
-      void do_add_request(RequestRef&& request,
-			  const C& client_id,
-			  const ReqParams& req_params,
-			  const Time time,
-			  const Cost cost = 1u) {
+      int do_add_request(RequestRef&& request,
+			 const C& client_id,
+			 const ReqParams& req_params,
+			 const Time time,
+			 const Cost cost = 1u) {
 	++tick;
 
         auto insert = client_map.emplace(client_id, ClientRecRef{});
@@ -944,7 +944,8 @@ namespace crimson {
 #if USE_PROP_HEAP
 	prop_heap.adjust(client);
 #endif
-      } // add_request
+	return 0;
+      } // do_add_request
 
       // data_mtx must be held by caller
       void update_next_tag(DelayedTagCalc delayed, ClientRec& top,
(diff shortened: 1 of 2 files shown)


src/dmclock_server.h
```diff
@@ -771,10 +776,10 @@ namespace crimson {
 			std::chrono::duration<Rep,Per> _idle_age,
 			std::chrono::duration<Rep,Per> _erase_age,
 			std::chrono::duration<Rep,Per> _check_time,
-			bool _allow_limit_break,
+			AtLimit _at_limit,
```
ivancich: What about another constructor that would invoke this one, but would take the bool as the fourth parameter and map true/false into Allow/Wait. That way old code would still work.

src/dmclock_server.h
```diff
@@ -804,36 +809,29 @@ namespace crimson {
 
 
       // data_mtx must be held by caller
-      void do_add_request(RequestRef&& request,
-			  const C& client_id,
-			  const ReqParams& req_params,
-			  const Time time,
-			  const double cost = 0.0) {
+      int do_add_request(RequestRef&& request,
```
ivancich: add documentation for all possible return values in comment above

cbodley:
One of the drawbacks of `AtLimit::Reject`/`Allow` as opposed to `AtLimit::Wait` is that it can cause the queues to run empty, which reduces the effectiveness of weighting between clients.

We discussed the idea of a 'reject threshold', such that `add_request()` would only return `EAGAIN` when `tag.limit > time + threshold`. I think this would be a good way to balance the queue depth with the desire to enforce an upper bound on the time spent waiting in the queue (and the memory use associated with queued requests).

theanalyst:
> We discussed the idea of a 'reject threshold', such that add_request() would only return EAGAIN when tag.limit > time + threshold. I think this would be a good way to balance the queue depth with the desire to enforce an upper bound on the time spent waiting in the queue (and the memory use associated with queued requests).

Is the threshold going to be per client or more at a queue level?

src/dmclock_server.h
```diff
@@ -1497,57 +1509,58 @@ namespace crimson {
...
-			      client_id,
-			      req_params,
-			      time,
-			      cost);
+	int r = super::do_add_request(std::move(request),
+				      client_id,
+				      req_params,
+				      time,
+				      cost);
 	schedule_request();
```
theanalyst: should schedule still be called if the negative return case?

src/dmclock_server.h
```diff
@@ -791,6 +798,8 @@ namespace crimson {
       {
 	assert(_erase_age >= _idle_age);
 	assert(_check_time < _idle_age);
+	// AtLimit::Reject depends on ImmediateTagCalc
+	assert(at_limit != AtLimit::Reject || !IsDelayed);
```
theanalyst: wouldn't this mean using PullPQ where isDelayed is set would assert on construction if we set AtLimit?

cbodley: yes, because AtLimit::Reject is not compatible with DelayedTagCalc

in order to enforce AtLimit::Reject, do_add_request() compares the request time against RequestTag::limit - but with DelayedTagCalc, we may start with an empty RequestTag, and only calculate a real one when the request gets to the top of the `ClientRec::requests` queue

src/dmclock_server.h
theanalyst: I guess once this is addressed, this pr is good to go?

ivancich: In discussions with Casey, I no longer think having another version of the constructor would help. Users of the library would have to make changes to use this version of the library.

src/dmclock_server.h
ivancich: So I'm thinking this should be addressed before we merge this.

cbodley:
i'll need to add some tests for the threshold piece

ivancich:
@cbodley OK!

src/dmclock_server.h
```diff
@@ -779,10 +804,11 @@ namespace crimson {
 			std::chrono::duration<Rep,Per> _idle_age,
 			std::chrono::duration<Rep,Per> _erase_age,
 			std::chrono::duration<Rep,Per> _check_time,
-			bool _allow_limit_break,
+			AtLimitParam at_limit_param,
 			double _anticipation_timeout) :
 	client_info_f(_client_info_f),
-	allow_limit_break(_allow_limit_break),
+	at_limit(get_or_default<AtLimit>(at_limit_param, AtLimit::Reject)),
```
theanalyst: should we default to AtLimit::Reject?

ivancich: The standard dmclock algorithm uses AtLimit::Wait. So if there's a default, I think that should be it. We're trying to keep the library as generic as possible.

Also AtLimit::Reject puts additional requirements on the caller that neither Wait or Allow require (i.e., the caller has to do something different depending on whether the request is rejected or not). Additionally at this point we have so little experience with AtLimit::Reject.

cbodley: `Reject` is only chosen as a default when the type of the `at_limit_param` is not `AtLimit`, so it must be `RejectThreshold`

at a higher level, the Pull/Push queue constructors both provide a default argument for this:
`AtLimitParam at_limit_param = AtLimit::Wait`

that maintains the same behavior as the previous default:
`bool _allow_limit_break = false`

cbodley:
updated with a `pull_reject_threshold` test

PR #51, opened 2018-03-27
server: delayed tag calculation as a template parameter

cbodley:
part of ongoing work in https://github.com/ceph/dmclock/pull/50, where `AtLimit::Reject` is not compatible with delayed tag calculation. posting separately for review

provides two overloads of functions `initial_request_tag()`, `update_next_tag()`, and `reduce_reservation_tags()`, and uses [tag dispatch](https://www.boost.org/community/generic_programming.html#tag_dispatching) to select the desired overload

### DIFF
--- a/src/dmclock_server.h
+++ b/src/dmclock_server.h
@@ -17,13 +17,6 @@
 
 /* COMPILATION OPTIONS
  *
- * By default we include an optimization over the originally published
- * dmclock algorithm using not the values of rho and delta that were
- * sent in with a request but instead the most recent rho and delta
- * values from the requests's client. To restore the algorithm's
- * original behavior, define DO_NOT_DELAY_TAG_CALC (i.e., compiler
- * argument -DDO_NOT_DELAY_TAG_CALC).
- *
  * The prop_heap does not seem to be necessary. The only thing it
  * would help with is quickly finding the mininum proportion/prioity
  * when an idle client became active. To have the code maintain the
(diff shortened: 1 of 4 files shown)


src/dmclock_server.h
```diff
@@ -989,10 +983,10 @@ namespace crimson {
 	for (auto& r : client.requests) {
 	  r.tag.reservation -= client.info->reservation_inv;
 
-#ifndef DO_NOT_DELAY_TAG_CALC
-	  // reduce only for front tag. because next tags' value are invalid
-	  break;
-#endif
+	  if (tag_calc == TagCalc::Delayed) {
```
ivancich: Would it be better to flip the nexting (word?) of the r-loop and conditional, so we don't run the conditional each time through the loop? So out the outer level, if we're delayed, we simply adjust the reservation for the first request. But if we're immediate only then we execute the loop and adjust all reservation tags.

cbodley:
@ivancich thanks for review. i moved the tag_calc comparison out of the loop, and added the missing TagCalc argument in sim/src/test_dmclock_main.cc

ivancich:
LGTM!

ivancich:
@cbodley I apologize. My PR made this merge not automatic. Sorry to have to ask you to rebase.

cbodley:
hi @ivancich, i've updated this branch to use tag dispatch on a template argument

ivancich:
Very nice change implementation. Thanks! I hereby reaffirm my approval.

PR #52, opened 2018-03-28
 Added const references to various function parameters

Rubonnek:
This pull request adds a small enhancement by avoiding copying data when unnecessary and mitigates the following cppcheck messages:

```
[sim/src/ConfUtils.cc:41]: (performance) Function parameter 'val_' should be passed by reference.
[sim/src/ConfUtils.cc:42]: (performance) Function parameter 'newsection_' should be passed by reference.
[sim/src/ConfUtils.cc:42]: (performance) Function parameter 'comment_' should be passed by reference.
[sim/src/ConfUtils.h:40]: (performance) Function parameter 'val_' should be passed by reference.
[sim/src/ConfUtils.h:41]: (performance) Function parameter 'newsection_' should be passed by reference.
[sim/src/ConfUtils.h:41]: (performance) Function parameter 'comment_' should be passed by reference.
[sim/src/str_list.h:96]: (performance) Function parameter 'sep' should be passed by reference.
[sim/src/simulate.h:308]: (performance) Function parameter 'time_unit' should be passed by reference.
[sim/src/simulate.h:354]: (performance) Function parameter 'time_unit' should be passed by reference.
[support/src/run_every.h:55]: (performance) Function parameter '_body' should be passed by reference.
```

### DIFF
--- a/support/src/run_every.h
+++ b/support/src/run_every.h
@@ -52,7 +52,7 @@ namespace crimson {
 
     template<typename D>
     RunEvery(D                     _wait_period,
-	     std::function<void()> _body) :
+	     const std::function<void()>& _body) :
       wait_period(duration_cast<milliseconds>(_wait_period)),
       body(_body)
     {
--- a/sim/src/ConfUtils.cc
+++ b/sim/src/ConfUtils.cc
@@ -38,8 +38,8 @@ using std::string;
 
 ////////////////////////////// ConfLine //////////////////////////////
 ConfLine::
-ConfLine(const std::string &key_, const std::string val_,
-      const std::string newsection_, const std::string comment_, int line_no_)
+ConfLine(const std::string &key_, const std::string &val_,
+      const std::string &newsection_, const std::string &comment_, int line_no_)
   : key(key_), val(val_), newsection(newsection_)
 {
   // If you want to implement writable ConfFile support, you'll need to save
--- a/sim/src/ConfUtils.h
+++ b/sim/src/ConfUtils.h
@@ -37,8 +37,8 @@
  */
 class ConfLine {
 public:
-  ConfLine(const std::string &key_, const std::string val_,
-	   const std::string newsection_, const std::string comment_, int line_no_);
+  ConfLine(const std::string &key_, const std::string &val_,
+	   const std::string &newsection_, const std::string &comment_, int line_no_);
   bool operator<(const ConfLine &rhs) const;
   friend std::ostream &operator<<(std::ostream& oss, const ConfLine &l);
 
--- a/sim/src/simulate.h
+++ b/sim/src/simulate.h
@@ -305,7 +305,7 @@ namespace crimson {
 
       template<typename T>
       void display_server_internal_stats(std::ostream& out,
-					 std::string time_unit) {
+					 const std::string& time_unit) {
 	T add_request_time(0);
 	T request_complete_time(0);
 	uint32_t add_request_count = 0;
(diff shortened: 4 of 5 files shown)


sim/src/sim_client.h
```diff
@@ -149,7 +149,7 @@ namespace crimson {
 
     public:
 
-      SimulatedClient(ClientId _id,
+      SimulatedClient(const ClientId& _id,
```
chardan: If these "Id" types are actually just ints, you're probably better off passing them as const values rather than const references.

support/src/run_every.h
```diff
@@ -52,7 +52,7 @@ namespace crimson {
 
     template<typename D>
     RunEvery(D                     _wait_period,
-	     std::function<void()> _body) :
+	     const std::function<void()>& _body) :
```
chardan: Same here with a function pointer. As a general rule if you're using small/built-in types (ints, chars, iterators, etc.) usually it's best to pass them by const value if you in effect want to make a copy.

sim/src/sim_client.h
Rubonnek: Thanks for pointing it out, but ClientId is [not a primitive](https://github.com/ceph/ceph/blob/2739b6f1c52a45aa22a62313d140d769be532874/src/librbd/watcher/Types.h#L19-L47) type.

chardan: No problem-- the other thing I should point out after I hem and haw with myself is that with const& in tricker situations (like here, where I wrongly assumed it was as per the definition at "https://github.com/Rubonnek/dmclock/blob/41d97df5493b23bcbb8c3aaf53934231c82e48e4/sim/src/sim_recs.h"), any errors will be on the side of good. ;-)

support/src/run_every.h
Rubonnek: I understand, but std::function is not the same as a function pointer as far as I know. They are similar indeed, but a pointer cannot hold data other than the address it points to. An std::function can copy the function which this const reference will avoid if its not necessary.

sim/src/sim_client.h
Rubonnek: Welp, now I think you are right. I first saw the ClientId when I scanned the whole ceph repository with `cppcheck` which included this dmclock repository. I thought ClientId was also defined as a struct here, but it's not, it's just what you pointed out.

I also didn't notice the `inconclusive` flag from `cppcheck`. That usually brings up false positives and this change is included with that flag.

Thanks for the review! I'll drop these changes, and the other inconclusive ones.

support/src/run_every.h
chardan: std::function<> is a bit of a magical beast, as you're right, it's not the same as a function pointer. Instead, it's wrapping a Callable. I looked into this in more detail, and the answer is interesting: the code generated for a function pointer or reference-to-lambda is impressively concise. So the real question is "what happens when we copy a std::function that's pointing to some big Callable object" and the answer is... it will have to do the right thing and make a copy. Which could be expensive. 
So, one of two things needs to happen in that case: either the object needs to be moved, or the cost of that copy is something that's expected. So, you're right: if you want to avoid that happening at all, passing by const& looks like the way to go! Thanks for giving me the opportunity to look into it more.

sim/src/sim_client.h
chardan: Glad to be of help!

support/src/run_every.h
Rubonnek: Anytime! And thanks for the explanation! I was going to do that myself since I really didn't recall the details.

PR #54, opened 2018-04-18
[RFC] server: priority queue no longer wraps R in unique_ptr

cbodley:
let the application choose the representation of R, as long as it can be moved or copied. if unique_ptr is needed, then unique_ptr\<T\> can be used in the template parameter R

### DIFF
--- a/src/dmclock_server.h
+++ b/src/dmclock_server.h
@@ -244,10 +244,6 @@ namespace crimson {
       // we don't want to include gtest.h just for FRIEND_TEST
       friend class dmclock_server_client_idle_erase_Test;
 
-    public:
-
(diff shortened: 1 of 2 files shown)


ivancich:
I like the core change. I would like to see a bit documentation in comments explaining the three variants imagined -- moveable, copyable, unique_ptr<> -- providing a mionimal sense of the implications of each. Are copies/moves only done when a request enters/exits the queue. Or are copies/moves done while other ops are done on the queue. That would help users of the library decide what to use as the request.

A number of overloaded `add_request`s were also removed. And unless I'm mistaken, the time now has to be passed in. What is the thought process behind those changes?

cbodley:
may revisit in the future

PR #55, opened 2018-04-25
misc: mark constructors as explicit

dalgaaf:
Set 4 constructors as explicit to avoid implicit usage.

Fix for cppcheck warning:
 Class has a constructor with 1 argument that is
 not explicit. Such constructors should in general be explicit for
 type safety reasons. Using the explicit keyword in the constructor
 means some mistakes when using the class can be avoided.

 For more information check:
  https://www.codeproject.com/Articles/28663/Explicit-Constructor-in-C

Signed-off-by: Danny Al-Gaaf <danny.al-gaaf@bisect.de>

### DIFF
--- a/sim/src/sim_recs.h
+++ b/sim/src/sim_recs.h
@@ -105,7 +105,7 @@ namespace crimson {
     struct TestResponse {
       uint32_t epoch;
 
-      TestResponse(uint32_t _epoch) :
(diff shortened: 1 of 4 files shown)


ivancich:
I have no real objection to this. But realize 3 of the 4 modifications were to constructors used strictly in the unit testing code.

PR #56, opened 2018-05-10
Set of functionality additions and small adjustments.

ivancich:
This commit contains a bunch of minor changes.

First it allows a ClientProfile to be updated with new reservation,
weight, and limit values by adding an update function.

Second it adds an ability to invoke callbacks when a ClientInfo object
is removed due to the idle timeout. Testing this functionality has
been added to the unit tests.

Third we add the ability to clear a heap to help with a more
controlled tear-down.

Finally, dmclock-servers "cleaning" has been renamed "idle erase" to
better indicate the role. Types and variable names have been adjusted
accordingly.

### DIFF
--- a/src/dmclock_server.h
+++ b/src/dmclock_server.h
@@ -98,17 +98,18 @@ namespace crimson {
       double limit_inv;
 
       // order parameters -- min, "normal", max
-      ClientInfo(double _reservation, double _weight, double _limit) :
-	reservation(_reservation),
-	weight(_weight),
-	limit(_limit),
-	reservation_inv(0.0 == reservation ? 0.0 : 1.0 / reservation),
-	weight_inv(     0.0 == weight      ? 0.0 : 1.0 / weight),
-	limit_inv(      0.0 == limit       ? 0.0 : 1.0 / limit)
-      {
-	// empty
+      ClientInfo(double _reservation, double _weight, double _limit) {
+	update(_reservation, _weight, _limit);
       }
 
+      inline void update(double _reservation, double _weight, double _limit) {
+	reservation = _reservation;
+	weight = _weight;
+	limit = _limit;
+	reservation_inv = (0.0 == reservation) ? 0.0 : 1.0 / reservation;
+	weight_inv = (0.0 == weight) ? 0.0 : 1.0 / weight;
+	limit_inv = (0.0 == limit) ? 0.0 : 1.0 / limit;
+      }
 
       friend std::ostream& operator<<(std::ostream& out,
 				      const ClientInfo& client) {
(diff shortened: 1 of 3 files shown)


src/dmclock_server.h
```diff
@@ -783,15 +787,34 @@ namespace crimson {
...
 	    new RunEvery(check_time,
-			 std::bind(&PriorityQueueBase::do_clean, this)));
+			 std::bind(&PriorityQueueBase::do_idle_erase, this)));
       }
 
 
       ~PriorityQueueBase() {
 	finishing = true;
+
+	DataGuard g(data_mtx);
```
cbodley: locking in a destructor is generally unnecessary, unless it's to synchronize with the shutdown of an internal thread. you don't need to provide mutual exclusion between the destructor and public functions, because it would already be undefined behavior for one thread to call functions on an object that's being destructed in another thread

src/dmclock_server.h
```diff
@@ -783,15 +787,34 @@ namespace crimson {
...
+	limit_heap.clear();
+#if USE_PROP_HEAP
+	prop_heap.clear();
+#endif
+	resv_heap.clear();
+
+	for (auto c = client_map.begin(); client_map.end() != c; /* empty */) {
+	  auto current = c++;
+	  for (auto l : idle_erase_listeners) {
+	    (*l)(current->second->client);
```
cbodley: what's the motivation for notifying these listeners on destruction?

src/dmclock_server.h
```diff
@@ -757,12 +759,14 @@ namespace crimson {
       Duration                  idle_age;
       Duration                  erase_age;
       Duration                  check_time;
-      std::deque<MarkPoint>     clean_mark_points;
+      std::deque<MarkPoint>     idle_erase_mark_points;
 
-      // NB: All threads declared at end, so they're destructed first!
+      using IdleEraseListener = std::function<void(const C&)>;
+      std::set<const IdleEraseListener*> idle_erase_listeners;
```
cbodley: it doesn't seem safe to store these functions by pointer - especially when they're passed to `add_idle_erase_listener()` as a const reference. that means it's perfectly valid to call with a lambda, which we'd store as a pointer to a temporary object:
```c++
  q.add_idle_erase_listener([] (const C& c) {});
```
it looks like it was done this way so you could implement `remove_idle_erase_listener()` - but is this kind of dynamic registration really necessary? consider treating this similar to `ClientInfoFunc`, and take a single optional callback as a constructor argument. if the caller really needs dynamic registration, they could implement it themselves inside their callback. does that make sense?

src/dmclock_server.h
ivancich: As I think about this further, not allowing the listener to be dynamic constrains the owner but likely makes more sense overall.

ivancich:
@cbodley Thanks for all your input. I reworked it and I think the code is much better now.

tchaikov:
@ivancich i will review this change later this week. but do you mind splitting this change into smaller commits? i can help this with your permission if you don't have enough bandwidth on this.

PR #57, opened 2018-05-11
DNM: dynamic client info testing issues, root cause not yet determined

ivancich:
For reasons not yet determined, the
dmclock_server_pull.dynamic_cli_info_f unit test fails when run under
valgrind. The counts do not come out as expected. Wondering if it is
the max operation between calculated tag and current time....

Also, fix some gtest testing messaging to be more accurate.

### DIFF
--- a/test/test_dmclock_server.cc
+++ b/test/test_dmclock_server.cc
@@ -744,6 +744,7 @@ namespace crimson {
     }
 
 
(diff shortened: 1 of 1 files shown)


tchaikov:
@ivancich i think #88 should fix the valgrind issue.

tchaikov:
superseded by #88

PR #59, opened 2018-08-07
sim: fix server_soft_limit can't work properly

Yan-waller:
server_soft_limit from test configuration file is bool type, it should change as the AtLimit has been introduced. Otherwise the sim test can't work.

Signed-off-by: yanjun <yan.jun8@zte.com.cn>


ivancich:
Thank you. I'm surprised that the simulator compiles without this fix!

PR #60, opened 2018-08-09
cmake: expose dmclock::dmclock with more properties and cleanups

tchaikov:


ivancich:
Looks good! Built stand-alone and in the ceph source tree.

PR #61, opened 2018-08-14
cmake: tell if we are included as subproject or not

tchaikov:
so the super-project can included dmclock's root directory. as the
"dmclock" target's target_include_directory() uses "$PROJECT_SOURCE_DIR"

Signed-off-by: Kefu Chai <tchaikov@gmail.com>


ivancich:
Looks good. Thanks @tchaikov !

PR #62, opened 2018-08-29
cmake: build gtest only if dmclock_TEST is ON

rjfd:
Signed-off-by: Ricardo Dias <rdias@suse.com>


ivancich:
This is good to add. I'm not sure the "bug" label is appropriate, though.

PR #67, opened 2018-12-26
server: fix request blocked caused by do_clean

Yan-waller:
If there are a large number of clients become old enough to erase in a short time, request may be blocked for a long time. since `do_clean` hold the `data_mtx` lock, other functions `pull_request` or `empty`  have to  wait.

Actually, in one of our testcase of ceph, it switch the `osd_op_queue` to `mclock_client`,  then we create 75, 000 rbd images in a short time, which brings huge clients. it result in OSD be marked down.
```
2018-12-20 20:16:17.266463 7f8bf21b8700  1 heartbeat_map is_healthy 'OSD::osd_op_tp thread 0x7f8bc310c700' had timed out after 60
2018-12-20 20:16:17.691459 7f8bcd921700  0 log_channel(cluster) log [WRN] : Monitor daemon marked osd.15 down, but it is still running
2018-12-20 20:16:17.691491 7f8bcd921700  0 log_channel(cluster) log [DBG] : map e638 wrongly marked me down at e638
```
the root cause is `do_clean()` hold the lock for more than a few minutes,  `OSD::osd_op_tp thread` has been blocked at [#L10464](https://github.com/ceph/ceph/blob/luminous/src/osd/OSD.cc#L10464).

This change try to improve `do_clean` in three ways:
- limit number of old clients to erase at a time.
- adjust check time to do clean more frequent.
- shorten idle/erase age to keep less clients.

However,  I have no idea of why the origin values of `check_time/idle_age/erase_age` at [#L1306](https://github.com/ceph/dmclock/blob/master/src/dmclock_server.h#L1306), so I'm not very sure is this adjustment reasonable.

@ivancich it might bother you, could you take a look at this when you back from festivals.
Wish you a happy Christmas holiday.

Signed-off-by: yanjun <yan.jun8@zte.com.cn>

### DIFF
--- a/src/dmclock_server.h
+++ b/src/dmclock_server.h
@@ -65,6 +65,11 @@ namespace crimson {
       std::numeric_limits<double>::lowest();
     constexpr uint tag_modulo = 1000000;
 
+    constexpr auto standard_idle_age  = std::chrono::seconds(300);
+    constexpr auto standard_erase_age = std::chrono::seconds(600);
+    constexpr auto standard_check_time = std::chrono::seconds(60);
+    constexpr auto aggressive_check_time = std::chrono::seconds(5);
+
     enum class AtLimit {
       // requests are delayed until the limit is restored
       Wait,
@@ -1303,9 +1308,9 @@ namespace crimson {
 			AtLimitParam at_limit_param = AtLimit::Wait,
 			double _anticipation_timeout = 0.0) :
 	PullPriorityQueue(_client_info_f,
-			  std::chrono::minutes(10),
-			  std::chrono::minutes(15),
-			  std::chrono::minutes(6),
+			  standard_idle_age,
+			  standard_erase_age,
+			  standard_check_time,
 			  at_limit_param,
 			  _anticipation_timeout)
       {
@@ -1542,9 +1547,9 @@ namespace crimson {
 	PushPriorityQueue(_client_info_f,
 			  _can_handle_f,
 			  _handle_f,
-			  std::chrono::minutes(10),
-			  std::chrono::minutes(15),
-			  std::chrono::minutes(6),
+			  standard_idle_age,
+			  standard_erase_age,
+			  standard_check_time,
 			  at_limit_param,
 			  _anticipation_timeout)
       {
--- a/src/dmclock_server.h
+++ b/src/dmclock_server.h
@@ -69,6 +69,7 @@ namespace crimson {
     constexpr auto standard_erase_age = std::chrono::seconds(600);
     constexpr auto standard_check_time = std::chrono::seconds(60);
     constexpr auto aggressive_check_time = std::chrono::seconds(5);
+    constexpr uint standard_erase_max = 100;
 
     enum class AtLimit {
       // requests are delayed until the limit is restored
@@ -811,6 +812,10 @@ namespace crimson {
       Duration                  erase_age;
       Duration                  check_time;
       std::deque<MarkPoint>     clean_mark_points;
+      // max number of clients to erase at a time
+      Counter erase_max;
+      // unfinished last erase point
+      Counter last_erase_point = 0;
 
       // NB: All threads declared at end, so they're destructed first!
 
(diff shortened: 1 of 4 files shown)


ivancich:
@Yan-waller Today's my first day back from our shutdown. I'll take a look at this.

Yan-waller:
@ivancich May I have your thoughts about this issue?

xiexingguo:
@ivancich ping

src/dmclock_server.h
```diff
@@ -1301,13 +1310,15 @@ namespace crimson {
...
       PullPriorityQueue(typename super::ClientInfoFunc _client_info_f,
 			AtLimitParam at_limit_param = AtLimit::Wait,
-			double _anticipation_timeout = 0.0) :
+			double _anticipation_timeout = 0.0,
+			Counter _erase_max = 100) :
 	PullPriorityQueue(_client_info_f,
-			  std::chrono::minutes(10),
-			  std::chrono::minutes(15),
-			  std::chrono::minutes(6),
+			  std::chrono::seconds(300),
```
ivancich: I know I'm responsible for this, but I'd prefer these constants not being scattered through the code and even repeated between PullPriorityQueue and PushPriorityQueue. What if we were to add some constexpr declarations at the top of this file (e.g., standard_idle_age, standard_erase_age, standard_check_time, and standard_erase_max) and use those below? Would you be alright with that?

src/dmclock_server.h
```diff
@@ -1522,10 +1533,11 @@ namespace crimson {
 			std::chrono::duration<Rep,Per> _erase_age,
 			std::chrono::duration<Rep,Per> _check_time,
 			AtLimitParam at_limit_param = AtLimit::Wait,
-			double anticipation_timeout = 0.0) :
+			double anticipation_timeout = 0.0,
+			Counter _erase_max = 100) :
```
ivancich: Do you foresee users wanting to change this? If not, why don't we move this to after _check_time and make it fixed as well. I realize that it will require coordinated changes with ceph itself when it updates the version of dmclock that it uses, but I'd rather be consistent than convenient.

src/dmclock_server.h
```diff
@@ -1301,13 +1310,15 @@ namespace crimson {
...
-			double _anticipation_timeout = 0.0) :
+			double _anticipation_timeout = 0.0,
+			Counter _erase_max = 100) :
 	PullPriorityQueue(_client_info_f,
-			  std::chrono::minutes(10),
-			  std::chrono::minutes(15),
-			  std::chrono::minutes(6),
+			  std::chrono::seconds(300),
+			  std::chrono::seconds(600),
+			  std::chrono::seconds(5),
```
ivancich: How did you arrive at 5 seconds? Did you do any profiling to see if this is good? Have you observed differences in behavior?

I'm thinking to your example of 75,000 rbd images. If you can only clean 100 at a time it will take 750 rounds to clean all of them. If those rounds are 1 minute apart, that's 750 minutes, which is 12.5 hours. If those rounds are 5 seconds apart then it's just a little over an hour. And that's a good argument.

But what about the case where no cleaning is to be done?

I can be convinced either way. Let me throw out an idea. What if there's a short check_time if you hit erase_max, say of 5 seconds. But if you didn't hit erase_max last time, you'll have a longer check_time.

I'd welcome your thoughts and feedback.

src/dmclock_server.h
Yan-waller: yes, I think it's quite reasonable to add constexpr declararions.

src/dmclock_server.h
Yan-waller: yeah, I originally put `_erase_max` here as an argument to make it convenient to change as needed, then I add a configurable item and pass it in from ceph, but I'm not realize that it brings compatibility problems since dmclock has been change from subtree to submodule in ceph (we are Luminous), so I'll drop it and make it fixed.

src/dmclock_server.h
Yan-waller: @ivancich thank you. I'm try to fix as above and repush it once I'm ready.

ivancich: Great! Thank you!

Yan-waller:
@ivancich fixed as our discussion above, could you take a look at it again?

We count the clients in queue and dump it every 5 secs, it works well as expected:
```
Fri Jan 18 17:45:44 CST 2019
39916
Fri Jan 18 17:45:51 CST 2019
39815
Fri Jan 18 17:45:57 CST 2019
39715
Fri Jan 18 17:46:04 CST 2019
39615
Fri Jan 18 17:46:10 CST 2019
39515
Fri Jan 18 17:46:16 CST 2019
39415
Fri Jan 18 17:46:23 CST 2019
39315
Fri Jan 18 17:46:29 CST 2019
39215
```

ivancich:
Very nice! Thank you!

PR #73, opened 2019-09-03
cmake: use parent C++ settings if possible

tchaikov:
Signed-off-by: Kefu Chai <tchaikov@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,8 +10,12 @@ endif()
 
 find_package(Threads)
 
-set(CMAKE_CXX_STANDARD 11)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
+if(CMAKE_CXX_STANDARD OR CMAKE_CXX_FLAGS MATCHES "-std=(c|gnu)\\+\\+")
+  # use existing settings if available
+else()
+  set(CMAKE_CXX_STANDARD 11)
+  set(CMAKE_CXX_STANDARD_REQUIRED ON)
+endif()
 
 add_subdirectory(src)
 


CMakeLists.txt
```diff
@@ -10,8 +10,19 @@ endif()
...
+  # standalone
+  set(CMAKE_CXX_STANDARD 11)
+  set(CMAKE_CXX_STANDARD_REQUIRED ON)
+else()
+  # embedded
+  if(CMAKE_CXX_STANDARD OR
+      CMAKE_CXX_FLAGS MATCHES "-std=c\\+\\+" OR
+      CMAKE_CXX_FLAGS MATCHES "-std=gnu\\+\\+")
+    # use parent project's settings
+  endif()
```
cbodley: shouldn't there be an else() case that does set CMAKE_CXX_STANDARD? maybe flip it around:
```cmake
# use parent project's settings if available
if(NOT CMAKE_CXX_STANDARD AND NOT CMAKE_CXX_FLAGS MATCHES "-std=")
  set(CMAKE_CXX_STANDARD 11)
  set(CMAKE_CXX_STANDARD_REQUIRED ON)
endif()
```

tchaikov: @cbodley yeah, i removed the check for standalone/embedded modes. as, in theory, use can also pass `CMAKE_CXX_FLAGS` and/or `CMAKE_CXX_STANDARD` by `cmake` command line. 

also, i still think

- it's easier to understand to avoid `NOT` operator
- `-std=` is not good enough. what if user passes options like `-std=gnu99`. it's probably pedantic. but `-std=c++17` and `-std=gnu++17` are more self-explained than `-std=`, IMHO.

also i consolidate these two regex into a single one, hopefully it's better.

PR #74, opened 2019-10-01
Small fixes to dmclock_server and run_every includes

athanatos:


athanatos:
Part of https://github.com/ceph/ceph/pull/30650

src/dmclock_server.h
```diff
@@ -42,7 +42,7 @@
 #include <boost/variant.hpp>
 
 #include "indirect_intrusive_heap.h"
-#include "run_every.h"
+#include "../support/src/run_every.h"
```
tchaikov: @athanatos what problem we want to address with this change? as we have

https://github.com/ceph/dmclock/blob/1c2a7d5491268c19ed5b4b38d9ae075420b4a0d5/src/CMakeLists.txt#L8-L10

so i assume `support/src/run_every.h` should be available when searching for `run_every.h`.

the reason i don't like this change is that it's unusual to have 
```c++
#include "../path/to/file.h
```
in a public header file exposed by a library. as an alternative, probably we can promote `run_every.h` and `run_every.cc`, so they live in `src/`?

athanatos: I actually just mirrored the way dmclock_client.h already works.  I'd rather put off cleaning up the directory structure for another patch.

PR #75, opened 2019-12-20
indirect_intrusive_heap: fix vector oob reference when removing greatest element

athanatos:


athanatos:
http://pulpito.ceph.com/sjust-2019-12-20_10:21:53-rados-wip-sjust-dmclock-fix-master-distro-basic-smithi

Above run filtered for centos 8 -- other failures appear unrelated/similar to master runs.

ivancich:
Thanks for these fixes, @athanatos.

PR #76, opened 2020-04-08
Make K use a standard type

PhoenixMage:
uint is undefined on musl as it isnt a standard type.
Replaced uint with the standard type "unsigned int" to resolve compilation issue on musl.

### DIFF
--- a/support/src/indirect_intrusive_heap.h
+++ b/support/src/indirect_intrusive_heap.h
@@ -48,7 +48,7 @@ namespace crimson {
 	   typename T,
 	   IndIntruHeapData T::*heap_info,
 	   typename C,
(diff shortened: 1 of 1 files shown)


support/src/indirect_intrusive_heap.h
```diff
@@ -48,7 +48,7 @@ namespace crimson {
 	   typename T,
 	   IndIntruHeapData T::*heap_info,
 	   typename C,
-	   uint K = 2>
+	   unsigned int K = 2>
```
tchaikov, line 51: nit, might want to just put `unsigned` instead of `unsigned int` for the sake of consistency.

PR #83, opened 2020-09-24
Test: Add test to verify requests are not scheduled if limit is exceeded

sseshasa:
Add a test to verify that when a client exceeds its limit, the
scheduler does not schedule the requests from that client but instead
responds with a future request type until the limit is restored.

The motivation for adding this test is that this scenario is currently not
handled within the mClockScheduler::dequeue() function i.e. handle
future request type response. The code to handle this case is currently
being put in place and tested as part of another PR.

### DIFF
--- a/test/test_dmclock_server.cc
+++ b/test/test_dmclock_server.cc
@@ -1359,5 +1359,116 @@ namespace crimson {
       EXPECT_EQ(0, pq.add_request_time({}, client1, {}, Time{3})); // 3 over
     }
 
+
+    TEST(dmclock_server_pull, pull_wait_at_limit) {
+      using ClientId = int;
+      using Queue = dmc::PullPriorityQueue<ClientId,Request>;
+      using QueueRef = std::unique_ptr<Queue>;
+
+      ClientId client1 = 52;
+      ClientId client2 = 8;
+
+      // Create client1 with high limit.
+      // Create client2 with low limit and with lower weight than client1
+      dmc::ClientInfo info1(1.0, 2.0, 100.0);
+      dmc::ClientInfo info2(1.0, 1.0, 2.0);
+
+      auto client_info_f = [&] (ClientId c) -> const dmc::ClientInfo* {
+        if (client1 == c) {
+          return &info1;
+        } else if (client2 == c) {
+          return &info2;
+        } else {
+          ADD_FAILURE() << "client info looked up for non-existent client";
+          return nullptr;
+        }
+      };
+
+      QueueRef pq(new Queue(client_info_f, AtLimit::Wait));
+
+      ReqParams req_params(1,1);
+
+      // make sure all times are before now
+      auto add_time = dmc::get_time() - 1.0;
+      auto old_time = add_time;
+
+      for (int i = 0; i < 50; ++i) {
+        EXPECT_EQ(0, pq->add_request_time(Request{}, client1, req_params, add_time));
+        EXPECT_EQ(0, pq->add_request_time(Request{}, client2, req_params, add_time));
+        add_time += 0.01;
+      }
+
+      EXPECT_EQ(2u, pq->client_count());
+      EXPECT_EQ(100u, pq->request_count());
+      int c1_count = 0;
+      int c2_count = 0;
+
+      // Pull couple of requests, should come from reservation queue.
+      // One request each from client1 and client2 should be pulled.
+      for (int i = 0; i < 2; ++i) {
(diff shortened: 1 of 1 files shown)


athanatos (approved):

test/test_dmclock_server.cc
```diff
@@ -1359,5 +1359,110 @@ namespace crimson {
...
+      ClientId client1 = 52;
+      ClientId client2 = 8;
+
+      // Create client1 with high limit.
+      // Create client2 with low limit and with lower weight than client1
+      dmc::ClientInfo info1(1.0, 2.0, 100.0);
+      dmc::ClientInfo info2(1.0, 1.0, 2.0);
+
+      auto client_info_f = [&] (ClientId c) -> const dmc::ClientInfo* {
+        if (client1 == c) return &info1;
```
ivancich, line 1377: Please use ceph coding style as documented at https://github.com/ceph/ceph/blob/master/CodingStyle .

test/test_dmclock_server.cc
```diff
@@ -1359,5 +1359,110 @@ namespace crimson {
...
+      for (int i = 0; i< 2; ++i) {
+        Queue::PullReq pr = pq->pull_request();
+        EXPECT_EQ(Queue::NextReqType::returning, pr.type);
+        auto& retn = boost::get<Queue::PullReq::Retn>(pr.data);
+
+        if (client1 == retn.client) {
+          ++c1_count;
+        } else if (client2 == retn.client) {
+          ++c2_count;
+        } else ADD_FAILURE() << "got request from neither of two clients";
```
ivancich, line 1415: Please use ceph coding style as documented at https://github.com/ceph/ceph/blob/master/CodingStyle .

test/test_dmclock_server.cc
```diff
@@ -1359,5 +1359,110 @@ namespace crimson {
...
+      for (int i = 0; i < 50; ++i) {
+        Queue::PullReq pr = pq->pull_request();
+        EXPECT_EQ(Queue::NextReqType::returning, pr.type);
+        auto& retn = boost::get<Queue::PullReq::Retn>(pr.data);
+
+        if (client1 == retn.client) {
+          ++c1_count;
+        } else if (client2 == retn.client) {
+          ++c2_count;
+        } else ADD_FAILURE() << "got request from neither of two clients";
```
ivancich, line 1440: Please use ceph coding style as documented at https://github.com/ceph/ceph/blob/master/CodingStyle .

test/test_dmclock_server.cc
```diff
@@ -1359,5 +1359,110 @@ namespace crimson {
...
+      }
+
+      EXPECT_EQ(2u, pq->client_count());
+      EXPECT_EQ(100u, pq->request_count());
+      int c1_count = 0;
+      int c2_count = 0;
+
+      // Pull couple of requests, should come from reservation queue.
+      // One request each from client1 and client2 should be pulled.
+      for (int i = 0; i< 2; ++i) {
```
ivancich, line 1406: missing space

ivancich (changes requested):
Looks good overall. Please use ceph coding style as documented at https://github.com/ceph/ceph/blob/master/CodingStyle . Otherwise, I approve.

test/test_dmclock_server.cc
sseshasa: Done.

test/test_dmclock_server.cc
sseshasa: Done.

test/test_dmclock_server.cc
sseshasa: Done.

test/test_dmclock_server.cc
sseshasa: Fixed.

sseshasa:
> Looks good overall. Please use ceph coding style as documented at https://github.com/ceph/ceph/blob/master/CodingStyle . Otherwise, I approve.

Thanks! There are other tests in the file where similar coding style needs to be adopted. That could be taken up as part of another PR.

ivancich:
> > Looks good overall. Please use ceph coding style as documented at https://github.com/ceph/ceph/blob/master/CodingStyle . Otherwise, I approve.
> 
> Thanks! There are other tests in the file where similar coding style needs to be adopted. That could be taken up as part of another PR.

Agreed, a clean-up PR is in order. And I'm probably responsible for some of the code that needs the clean-up.

ivancich (approved):

PR #84, opened 2020-11-05
server: fix run_sched_ahead() scheduling issue

yongseokoh:
The first issue is that the run_sched_ahead() function is blocked when the push priority queue is employed. It is confirmed that the sched_ahead_cv.wait() isn't signaled if the predicate function returns false with this->finishing variable.  For this reason, sched_ahead_when > TimeZero condition is added pred() function. 

The other is that the further request is not fetched when the schedule_request() executes the submit_request() and  sched_ahead_when is set to TimeZero. It implies that run_sched_ahead thread can't wake up by itself. This commit adds that the schedule_request() is called while requests are ready and sched_ahead_when is equal to TimeZero. A unit test has also been added to check whether 3 requests are dispatched after a few seconds.

Fixes: https://tracker.ceph.com/issues/48123
Fixes: https://tracker.ceph.com/issues/48317
Signed-off-by: Yongseok Oh <yongseok.oh@linecorp.com>

### DIFF
--- a/src/dmclock_server.h
+++ b/src/dmclock_server.h
@@ -1639,7 +1639,7 @@ namespace crimson {
 				      time,
 				      cost);
         if (r == 0) {
-	  schedule_request();
+	  (void) schedule_request();
         }
 #ifdef PROFILE
 	add_request_timer.stop();
@@ -1653,7 +1653,7 @@ namespace crimson {
 #ifdef PROFILE
 	request_complete_timer.start();
 #endif
-	schedule_request();
+	(void) schedule_request();
 #ifdef PROFILE
 	request_complete_timer.stop();
 #endif
@@ -1738,11 +1738,11 @@ namespace crimson {
 
 
       // data_mtx should be held when called
-      void schedule_request() {
+      typename super::NextReqType schedule_request() {
 	typename super::NextReq next_req = next_request();
 	switch (next_req.type) {
 	case super::NextReqType::none:
-	  return;
+	  break;
 	case super::NextReqType::future:
 	  sched_at(next_req.when_ready);
 	  break;
@@ -1752,6 +1752,7 @@ namespace crimson {
 	default:
 	  assert(false);
 	}
+	return next_req.type;
       }
 
 
@@ -1777,9 +1778,12 @@ namespace crimson {
 
 	    l.unlock();
 	    if (!this->finishing) {
-	      typename super::DataGuard g(this->data_mtx);
-	      schedule_request();
-	    }
+	      do {
+ 	        typename super::DataGuard g(this->data_mtx);
+ 	        if (schedule_request() == super::NextReqType::future)
+ 	          break;
+	      } while (!this->empty());
+ 	    }
 	    l.lock();
 	  }
 	}
--- a/src/dmclock_server.h
+++ b/src/dmclock_server.h
@@ -1763,7 +1763,9 @@ namespace crimson {
 
 	while (!this->finishing) {
 	  // predicate for cond.wait()
-	  const auto pred = [this] () -> bool { return this->finishing; };
+	  const auto pred = [this] () -> bool {
+	    return this->finishing || sched_ahead_when > TimeZero;
+	  };
 
 	  if (TimeZero == sched_ahead_when) {
 	    sched_ahead_cv.wait(l, pred);
--- a/test/test_dmclock_server.cc
+++ b/test/test_dmclock_server.cc
@@ -270,6 +270,60 @@ namespace crimson {
     } // TEST
 
 
+    TEST(dmclock_server, schedule_req_pushprio_queue) {
+      struct MyReq {
+        int id;
+
+        MyReq(int _id) :
+          id(_id)
+        {
+          // empty
+        }
+      }; // MyReq
+
+      using ClientId = int;
+      using Queue = dmc::PushPriorityQueue<ClientId,MyReq>;
+      using MyReqRef = typename Queue::RequestRef;
+      ClientId client1 = 17;
+      ClientId client2 = 34;
+      ClientId client3 = 48;
+
+      dmc::ClientInfo ci(1.0, 1.0, 1.0);
+      auto client_info_f = [&] (ClientId c) -> const dmc::ClientInfo* {
+        return &ci;
+      };
+      auto server_ready_f = [] () -> bool { return true; };
+      auto submit_req_f = [] (const ClientId& c,
+                              std::unique_ptr<MyReq> req,
+                              dmc::PhaseType phase,
+                              uint64_t req_cost) {
+        // empty; do nothing
+      };
+
+      Queue pq(client_info_f,
+               server_ready_f,
+               submit_req_f,
+               std::chrono::seconds(3),
+               std::chrono::seconds(5),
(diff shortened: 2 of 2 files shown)


yongseokoh:
@cbodley @ivancich  Could you please take a look at this? We have been actively working on the MDS QoS [RFC](https://lists.ceph.io/hyperkitty/list/dev@ceph.io/message/FSUC4N7WLLMTUIZA4HYDPESASEKYV434/) based on the push priority queue. However, it sometimes is not working properly with this issue. We would like to confirm whether this PR needs to be considered.

ivancich:
@yongseokoh I will try to review this in the next day or so. It's been a while since I've been through this code, so I need to invest some time in it. Thanks for your contributions!

src/dmclock_server.h
```diff
@@ -1762,13 +1764,14 @@ namespace crimson {
 
 	while (!this->finishing) {
 	  // predicate for cond.wait()
-	  const auto pred = [this] () -> bool { return this->finishing; };
+	  const auto pred = [this] () -> bool { return this->finishing || this->notified; };
```
athanatos, line 1767: Why isn't this predicate simply this->finishing || sched_ahead_when != TimeZero?  Seems like adding notified is unnecessary.

athanatos: Also, I think you should combine the two PRs, reviewing these in isolation doesn't make a ton of sense as one modifies the cond variable predicate and the other modifies the loop behavior.

yongseokoh: @athanatos I think this->notified can be replaced with sched_ahead_when > TimeZero. I just confirmed  request_count() internally tries to hold the data_mtx lock, leading to dead lock issue. Is it okay?

I will merge https://github.com/ceph/dmclock/pull/86 with this PR.

yongseokoh: @athanatos I have made changes and cherry-picked #86 here.

athanatos:
@ivancich Have you had a chance to take a look?

yongseokoh:
@ivancich  Could you please take a look at this? I have made some efforts to this project based on @athanatos' review.

src/dmclock_server.h
```diff
@@ -1762,13 +1762,13 @@ namespace crimson {
...
 	  // predicate for cond.wait()
-	  const auto pred = [this] () -> bool { return this->finishing; };
+	  const auto pred = [this] () -> bool { return this->finishing || sched_ahead_when > TimeZero; };
 
 	  if (TimeZero == sched_ahead_when) {
 	    sched_ahead_cv.wait(l, pred);
 	  } else {
 	    // cast from Time -> duration<Time> -> Duration -> TimePoint
-	    const auto until = typename super::TimePoint{
+	    const auto until = typename std::chrono::system_clock::time_point{
```
tchaikov, line 1771: @yongseokoh could you elaborate a little bit why we need to have this change? i think any timepoint should work here as long as `chrono::is_clock_v<TimePoint::clock>. as the standard library should convert it to `steady_clock` or `system_clock`.

yongseokoh: @tchaikov that's a good point. I will revert the line.

src/dmclock_server.h
```diff
@@ -1777,8 +1777,14 @@ namespace crimson {
...
-	      typename super::DataGuard g(this->data_mtx);
-	      schedule_request();
+	      auto get_sched_ahead_when = [this] () -> Time {
+	        std::lock_guard<std::mutex> l(sched_ahead_mtx);
+	        return sched_ahead_when;
+	      };
+	      while (get_sched_ahead_when() == TimeZero && this->request_count() > 0) {
+	        typename super::DataGuard g(this->data_mtx);
+	        schedule_request();
+	      }
```
tchaikov, line 1787: i think the first call of `get_sched_ahead_when()` should always return `TimeZero`. so, for better performance, how about rewriting the `while` loop like

```c++
do {
  typename super::DataGuard g(this->data_mtx);
  schedule_request();
} while (get_sched_ahead_when() == TimeZero && has_request());
```

in comparison with the existing behavior, each loop always needs to acquire the `sched_ahead_mtx` lock one more time no matter  if the next request should be scheduled in future or not, i think we can improve the performance further by making `schedule_request()` return a boolean or an enum indicating if the next request should be scheduled right away or not. if it is, we should continue with the loop until hitting a request which will block the `run_sched_ahead()` thread.

tchaikov:
@yongseokoh sorry for not being able to send the comments in a single batch. i am still trying to get my head around the dmClock algorithm and our implementation.

src/dmclock_server.h
yongseokoh: @tchaikov based on your valuable review, I've modified some lines.

tchaikov (approved):

tchaikov:
@ivancich hi Eric, do you want to take a final look before this PR gets reviewed?

src/dmclock_server.h
```diff
@@ -1777,9 +1778,12 @@ namespace crimson {
...
 	    l.unlock();
 	    if (!this->finishing) {
-	      typename super::DataGuard g(this->data_mtx);
-	      schedule_request();
-	    }
+	      do {
+ 	        typename super::DataGuard g(this->data_mtx);
+ 	        if (schedule_request() == super::NextReqType::future)
+ 	          break;
+	      } while (this->request_count());
```
tchaikov, line 1785: nit,
```suggestion
	      } while (!empty());
```

src/dmclock_server.h
```diff
@@ -1752,6 +1752,7 @@ namespace crimson {
 	default:
 	  assert(false);
 	}
+        return next_req.type;
```
tchaikov, line 1755: nit, might want to use tab for indent to be consistent with the rest of this file

src/dmclock_server.h
```diff
@@ -1762,7 +1763,7 @@ namespace crimson {
 
 	while (!this->finishing) {
 	  // predicate for cond.wait()
-	  const auto pred = [this] () -> bool { return this->finishing; };
+	  const auto pred = [this] () -> bool { return this->finishing || sched_ahead_when > TimeZero; };
```
tchaikov, line 1766: nit, might want to put `return this->finishing || sched_ahead_when > TimeZero; };` at a new line for better readability, like
```c++
	  const auto pred = [this] () -> bool {
	    return this->finishing || sched_ahead_when > TimeZero;
	  };
```

src/dmclock_server.h
yongseokoh: Good suggestion. this pointer is necessary for empty().

src/dmclock_server.h
yongseokoh: I've adjusted the indentation.

src/dmclock_server.h
yongseokoh: The function has been reworked based on your feedback.

tchaikov:
@yongseokoh just one last thing, could you squash the fix-up commits into the original change ? for instance, to fold "server: replace request_count() with !empty() for better performance " into "server: allow schedule_request returns NextReqType that indicates us ..." ?

tchaikov (approved):

PR #85, opened 2020-11-12
util: Fix MinGW compilation error

aserdean:
MinGW complains:
```
/home/ubuntu/dmclock/src/dmclock_util.cc: In function ‘void crimson::dmclock::debugger()’:
/home/ubuntu/dmclock/src/dmclock_util.cc:33:9: error: ‘SIGCONT’ was not declared in this scope
   raise(SIGCONT);
         ^~~~~~~
/home/ubuntu/dmclock/src/dmclock_util.cc:33:9: note: suggested alternative: ‘SIGINT’
   raise(SIGCONT);
         ^~~~~~~
         SIGINT
```

SIGCONT is not defined on Windows.
Replace the call with `DebugBreak`.

Signed-off-by: Alin Gabriel Serdean <aserdean@cloudbasesolutions.com>

### DIFF
--- a/src/dmclock_util.cc
+++ b/src/dmclock_util.cc
@@ -30,5 +30,9 @@ std::string crimson::dmclock::format_time(const Time& time, unsigned modulo) {
 
 
 void crimson::dmclock::debugger() {
+#ifndef _WIN32
   raise(SIGCONT);
+#else
+  DebugBreak();
+#endif
 }


ivancich (approved):

ivancich (commented):
I did approve but now I'm questioning that decision. But now I think a conversation would be better....

The reason this function exists is so that when developing or checking existing code, it's very easy to drop into the debugger (e.g., gdb) by adding a call.

If we simply comment out the raising of the signal, then the function won't work as planned, in which case why even have the function.

So is there a Windows way to enter the debugger?

Of maybe in the Windows environment we should eliminate the function all together.

What are your thoughts?

aserdean:
No worries, I was hasty with the implementation, sorry.

I would like to keep the function because it may be used in other projects.

I have updated the commit and the message.

ivancich (approved):

PR #86, opened 2020-11-16
server: call schdule_request() in PushPriorityQueue::run_sched_ahead() while sched_ahead_when is TimeZero and requests are queued

yongseokoh:
If the schedule_request() function is called in the 1782 line and the request is serviced through submit_request(), sched_ahead_when must be configured with the next time to schedule correctly for the next request. However, in the current implementation, even though requests are available, sched_wait_wait is set to TimeZero. As a result, the next request must wait until request_completed() or add_request() is called. Also, a unit test has been added to check whether 3 requests have been serviced in the queue after 4 seconds.

Fixes: https://tracker.ceph.com/issues/48317
Signed-off-by: Yongseok Oh <yongseok.oh@linecorp.com>

### DIFF
--- a/src/dmclock_server.h
+++ b/src/dmclock_server.h
@@ -1777,8 +1777,10 @@ namespace crimson {
 
 	    l.unlock();
 	    if (!this->finishing) {
-	      typename super::DataGuard g(this->data_mtx);
-	      schedule_request();
+              while (sched_ahead_when == TimeZero && this->request_count() > 0) {
+                typename super::DataGuard g(this->data_mtx);
+                schedule_request();
+              }
 	    }
 	    l.lock();
 	  }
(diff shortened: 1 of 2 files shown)


yongseokoh:
@ivancich  Could you please assign reviewers?

ivancich:
@yongseokoh I will try to review this in the next day or so. It's been a while since I've been through this code, so I need to invest some time in it. Thanks for your contributions!

src/dmclock_server.h
```diff
@@ -1777,8 +1777,10 @@ namespace crimson {
 
 	    l.unlock();
 	    if (!this->finishing) {
-	      typename super::DataGuard g(this->data_mtx);
-	      schedule_request();
+              while (sched_ahead_when == TimeZero && this->request_count() > 0) {
```
athanatos, line 1780: Is it actually valid to access sched_ahead_when outside of sched_ahead_mtx?

yongseokoh:
This PR will be integrated into https://github.com/ceph/dmclock/pull/84. I will close this PR.

PR #92, opened 2024-11-19
Implement dynamic tag info functionality for accurate tag calculation with multiple queues configured per server

sseshasa:
To better understand the fix, the problem with multiple queues configured per server
needs to be understood first. This problem is currently faced by Ceph (or Ceph OSDs)
which are clients of dmClock server. Ceph creates multiple op queues per OSD (a.k.a
OSD op queue shards) and each op queue runs independently with the same server
id (OSD ID). With this configuration, tags are calculated on each queue independently.
Therefore, a client could distribute requests across the multiple mClock queues but
the server cannot meet the clients' QoS settings due to the following problem:

### Problem Description

Consider two queues configured on the same server with
items added in the following sequence:
    
       Enqueue--->|T5|T2|T0|--->Dequeue
                       Queue0
    
       Enqueue--->|T4|T3|T1|--->Dequeue
                       Queue1
    
Consider the request arrival times (_arr_time_) in each queue according
to the number associated with each tag, for example:
    
     - Req0 arrives on queue0 at time t0
     - Req1 arrives on queue1 at time t1
     - Req2 arrives on queue0 at time t2
and so on with the final Req5 arriving on queue0 at time t5

Consider x to be any of the  client info parameters i.e. reservation, weight or limit
    
  |  Tag calcs on Queue0    | Tag calcs on Queue1 |
  | ------------------------------------|---------------------------------|
  | T0 = max((NO_TAG + (1/x)), arr_time)| T1 = max((NO_TAG + (1/x)), arr_time)|
  | T2 = max((T0 + (1/x)), arr_time)    | T3 = max((T1 + (1/x)), arr_time)|
  |  T5 = max((T2 + (1/x)), arr_time)    | T4 = max((T3 + (1/x)), arr_time)|
    
For simplicity, assume reservation and limit is set to 3 IOPS, `AtLimit` is
set to '`Wait`' and all requests arrived within a few milliseconds on
both queues.
    
It's pretty clear that all the 3 requests from both the queues will be
scheduled resulting in 6 IOPS which is not correct.

### Proposed Solution

The fix is better understood by applying the solution to the above problem.
The fix is applicable to both `DelayedTagCalc` and `ImmediateTagCalc`.

**Pre-conditions:**

- Dynamic tag info functionality (bool U2 in the pull constructor) is enabled (`is_dynamic_tag_info_f`).
- The dmClock client maintains a mapping of `clientId` to `ReqTagInfo` for each client.
- The dmClock client registers custom implementations for `reqtag_updt_f` and  `reqtag_info_f`.
    
**Solution:**

The fix involves enabling the dynamic tag info for two operations by the following entities:
1.  _the server_ - to update (using `reqtag_updt_f`) the dmClock client with  latest tag for a `clientId` on a given queue once it's calculated. This  is called after tags are calculated in `initial_tag()` and ` update_next_tag()`.
    
2.  _the dmClock client_ - to calculate and update the tick interval before  adding a request to the mClock queue. For a given queue, the tick interval is the number of request(s) added to other mClock queues  _before_ the current one arriving on itself. The latest tick interval  is  read by the server using `reqtag_info_f` before calculating the ` initial_tag()`. See `calc_interval_tag()` on how the new tag is  calculated and see the unit tests on how the client calculates the  tick interval before adding the request to the queue. It's important to note that for `DelayedTagCalc`, the tag will be calculated as part of `initial_tag()` ONLY for non-zero tick intervals.
    
The tag calculation for each tag is shown below with the fix in place:
_(Apologies for the table formatting. Please consider the tag calculation row as the start of a row)_    
|Tag calcs on Queue0                 | Tag calcs on Queue1                      |
|------------------------------------|------------------------------------------|
|**T0 = max((NO_TAG + (1/x)), arr_time)**| **T1 = max((T0 + (i/x)), arr_time)**         |
|tick interval i = 0;                | tick interval i = 1 because req1         |
|                                    | arrived after req0 on Queue0.            |
|                                    | [Note: If tick_inteval > 0, the tag      |
|                                    |  is calculated as part of                |
|                                    |  initial_tag() for DelayedTagCalc]       |
|                                    |                                          |
|**T2 = max((T1 + (i/x)), arr_time)**    |**T3 = max((T2 + (i/x)), arr_time)**          |
|tick interval i = 1                 |tick interval i = 1                       |
|                                    |                                          |
|**T5 = max((T3 + (i/x)), arr_time)**   |**T4 = max((T3 + (1/x)), arr_time)**          |
|tick interval i = 2 because req5    |tick interval i = 0 because req4          |
|arrived two requests after T2.      |follows immediately after req3 on         |
|                                    |the same queue.                           |
|                                    |[**Note**: If tick_interval == 0, the tag      |
|                                    | is calculated only when it gets to the front|
|                                    |as part of `update_next_tag()` as usual |
|                                    |if DelayedTagCalc is enabled.]            |

For simplicity, assume reservation and limit is set to 3 IOPS, `AtLimit` is set to '`Wait`' and all requests arrived within a second on both queues.

With the fix in place only T0, T1 and T2 will be scheduled as expected during the first phase since each request is spaced` 1/res` apart i.e.,  1/3rd of a second. The same pattern will apply for the rest of the requests in the queues thus ensuring accurate scheduling of requests  across all the queues with the same server.
    
**Tests:**
 A lot of unit tests have been added to exercise the solution with both delayed and immediate tag calculation. Of particular importance are the  tests that completely randomize queue additions and pulls from multiple(5 Nos.)  queues. The tests also provide an insight into how the clients calculate the latest tick interval and how the latest tag is updated by the client via the dynamic functions. See:
     - pull_reservation_randomize_delydtag
     - pull_reservation_randomize_immtag
     - pull_weight_randomize_delydtag
     - pull_weight_randomize_immtag

Signed-off-by: Sridhar Seshasayee <sseshasa@redhat.com>

### DIFF
--- a/src/dmclock_recs.h
+++ b/src/dmclock_recs.h
@@ -18,6 +18,7 @@
 
 #include <ostream>
 #include <assert.h>
+#include <cstdint>
 
 
 namespace crimson {
--- a/src/dmclock_server.h
+++ b/src/dmclock_server.h
@@ -325,8 +325,9 @@ namespace crimson {
     //   originally published dmclock algorithm, allowing it to use the most
     //   recent values of rho and delta.
     // U1 determines whether to use client information function dynamically,
+    // U2 determines whether to use client tag information function dynamically
     // B is heap branching factor
-    template<typename C, typename R, bool IsDelayed, bool U1, unsigned B>
+    template<typename C, typename R, bool IsDelayed, bool U1, bool U2, unsigned B>
     class PriorityQueueBase {
       // we don't want to include gtest.h just for FRIEND_TEST
       friend class dmclock_server_client_idle_erase_Test;
@@ -399,7 +400,7 @@ namespace crimson {
       // ClientRec could be "protected" with no issue. [See comments
       // associated with function submit_top_request.]
       class ClientRec {
-	friend PriorityQueueBase<C,R,IsDelayed,U1,B>;
+	friend PriorityQueueBase<C,R,IsDelayed,U1,U2,B>;
 
 	C                     client;
 	RequestTag            prev_tag;
@@ -587,6 +588,12 @@ namespace crimson {
       // a function that can be called to look up client information
       using ClientInfoFunc = std::function<const ClientInfo*(const C&)>;
 
+      // a function that can be called to look up client tag information
+      using ReqTagInfoFunc = std::function<const ReqTagInfo*(const C&)>;
+
+      // a function that can be called to update client tag information
+      using ReqTagUpdtFunc = std::function<void(const C&,
+                                                const RequestTag&)>;
 
       bool empty() const {
 	DataGuard g(data_mtx);
@@ -803,7 +810,10 @@ namespace crimson {
       };
 
       ClientInfoFunc        client_info_f;
+      ReqTagInfoFunc        reqtag_info_f;
+      ReqTagUpdtFunc        reqtag_updt_f;
       static constexpr bool is_dynamic_cli_info_f = U1;
+      static constexpr bool is_dynamic_tag_info_f = U2;
 
 #ifdef WITH_SEASTAR
       static constexpr int data_mtx = 0;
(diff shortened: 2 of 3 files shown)


sseshasa:
@athanatos @ivancich @rzarzynski @neha-ojha Please take time to review this PR. I know most of you will be out for Cephalocon. In the meantime, I am hoping to get the changes on the Ceph side ready to get some real tests going. Thanks!

athanatos:
In your first example above, why can't we avoid the problem by simply setting per-queue reservation/limit to 1/3 (1/<num_queues>) of the value we want for the whole OSD?

athanatos:
reqtag_updt_f would be updating memory shared between queues, right?

sseshasa:
> In your first example above, why can't we avoid the problem by simply setting per-queue reservation/limit to 1/3 (1/<num_queues>) of the value we want for the whole OSD?

This may not work in all scenarios. For e.g., if a subset of queues are active for a workload, the realized IOPS may be lower than what is set by the client.

I recall having tried this on a cluster a while ago but don't remember what the result looked like. I will try it again and get back.

But I tried this with the randomized unit test without dynamic tag info and the number of requests dequeued from the client was lower than expected but not by much. I suspect on an actual cluster we may observe lower than expected IOPS.

Let me run a few tests on a cluster and get back with more details.

sseshasa:
> reqtag_updt_f would be updating memory shared between queues, right?

Yes, that's correct.

sseshasa:
> Let me run a few tests on a cluster and get back with more details.

With a single client, I ran a test using rados bench with your suggestion i.e., divide res and lim by number of shards. The results show that client is unable to achieve the set limit. Res was set to 125 IOPS and Limit to 625 IOPS and wgt set to 1. On each shard, res and lim was divided by the number of shards. For comparison, I ran the test with a single shard. Here's the outcome:


**client QoS:  [res:125 wgt:1 lim:625]**


**With 5 shards & res and limit reduced by num_shards**

```
Total time run:         120.346
Total writes made:      13405
Write size:             4096
Object size:            4096
Bandwidth (MB/sec):     0.435106
Stddev Bandwidth:       0.0268858
Max bandwidth (MB/sec): 0.480469
Min bandwidth (MB/sec): 0.359375
Average IOPS:           111
Stddev IOPS:            6.88277
Max IOPS:               123
Min IOPS:               92
Average Latency(s):     0.143412
Stddev Latency(s):      0.12554
Max latency(s):         0.640022
Min latency(s):         0.000296842
```


**With 1 OSD shard**

```
Total time run:         120.026
Total writes made:      74925
Write size:             4096
Object size:            4096
Bandwidth (MB/sec):     2.43845
Stddev Bandwidth:       0.0278175
Max bandwidth (MB/sec): 2.46875
Min bandwidth (MB/sec): 2.15625
Average IOPS:           624
Stddev IOPS:            7.12127
Max IOPS:               632
Min IOPS:               552
Average Latency(s):     0.0256237
Stddev Latency(s):      0.0040078
Max latency(s):         0.143611
Min latency(s):         0.000396659
```

rzarzynski:
@sseshasa, @athanatos: what's the status of this?

sseshasa:
Closing this PR as tests with corresponding changes on the client side (Ceph OSD & scheduler) did not show any significant improvement.

